PR #1, opened 2025-05-29
wip: git-use

cwlbraa:
warning: contains quite a bit of LLM generated code

this PR attempts to add git-centric functionality to the container-use mcp.

so far, it:
- propagates uncommitted state into the container, then commits it there
- correctly creates commits in the container for any action that writes to the workdir
- creates container-XYZ branches to track the history
- _attempts_ to propogate changes back to the host machine via `git bundle` and `git fetch bundlePath branchName:refs/remotes/branchName/main`, but fails
  - i attempted to do this with thin bundles (basically excluding what you've already got locally from the bundle) but it adds an extra variable to debugging the bundle unpack, so right now using `--all`, so we're sending the wholeass git repo over the wire each time.
- adds stderr tee debug printing so you can set `LOG_FILE=/tmp/cu.stderr.log` and see what's happening

i've tried a bunch of aggressive debugging techniques at this point, but with my current approach, tailing the err logs hangs right after (maybe in the middle of?) pushing bundles through File.Contents... the tools seem to continue onwards, but the stderr logging stops.


mcp.go
```diff
@@ -601,6 +636,67 @@ var ContainerFileDeleteTool = &Tool{
...
+	"remote_url": %q,
+	"captured_at": %q
+}`, gitState.RootPath, gitState.CurrentBranch, gitState.CurrentCommit, gitState.RemoteURL, 
+		gitState.CapturedAt.Format("2006-01-02 15:04:05"))
+		
+		return mcp.NewToolResultText(response), nil
+	},
+}
+
+var ContainerSyncTool = &Tool{
```
cwlbraa, line 631: this is for debugging only, intent was to build something that doesn't require manual sync at all

PR #4, opened 2025-06-03
Environment support

aluzzardi:


environment.go
```diff
@@ -0,0 +1,367 @@
...
+		}
+	}
+	return nil
+}
+
+type Environment struct {
+	ID           string `json:"id"`
+	Name         string `json:"name"`
+	Source       string `json:"-"`
+	Worktree     string `json:"-"`
```
cwlbraa, line 70: noice, i had been meaning to do this refactor, the functions were annoying

environment.go
```diff
@@ -0,0 +1,367 @@
...
+}
+
+var environments = map[string]*Environment{}
+
+func CreateEnvironment(ctx context.Context, explanation, source, name string) (*Environment, error) {
+	env := &Environment{
+		ID:           uuid.New().String(),
+		Name:         name,
+		Source:       source,
+		Dockerfile:   "FROM ubuntu:latest",
```
cwlbraa, line 88: conceptually, i'm a little wary of plumbing Dockerfile. On the one hand, I get that we're meeting LLMs where they're at, they know how to write dockerfiles. 

On the other hand, we could get away with `BaseImage: string`, and eventually that might mean leaning harder into the LLB-saving stuff we've talked about... 

Although layering source code maybe makes any LLB saving a bad fit for specifying and customizing base images. I'm kinda talking myself into Dockerfile lol

aluzzardi: Yeah, not sold on the idea, just needed a quick shortcut.

It's more than a `BaseImage`: it's that plus commands (e.g. `pip install`, manually installing a specific version of Go in Ubuntu, ...).

I reluctantly did that because:

- It's human readable/writeable
- LLMs are surprisingly good at writing them

I think it should be swapped out for something better (maybe just a `BaseImage` plus `SetupCommands`?) ... maybe as we add other keys (`services`, `secrets`, ...) it will make much sense. But for now, it's a gigantic shortcut

PR #8, opened 2025-06-05
Update makefile to move binary to users path

jasonmccallister:
This updates the `Makefile` to find a default writable directory in `$PATH` or falls back to another writeable path. 

It will build the binary as is using `docker` and then move the artifact into the users `$PATH`.


aluzzardi:
@jasonmccallister sorry, just renamed the binary to `cu` (was planning to since yesterday but postponed to not break things)

nit: in the spirit of makesfile, how about having `make` (builds binary locally, default) / `make install` (builds binary and moves to $PATH), and just saying `make install` in the README?

I would find it weird for `make` to touch files outside the project, but totally fine with `make install` doing that (also I can decide to only `make`)

jasonmccallister:
@aluzzardi absolutely, just merged your changes and will move that to install. Are you good with having the `install` run `cu` ahead of time?

jasonmccallister:
Ignore the comment about `cu` - needed to restart my shell :)

README.md
```diff
@@ -40,7 +40,7 @@ It's an open-source MCP server that works as a CLI tool with Claude Code, Cursor
 make
 ```
 
-Make sure to put `cu` in your `$PATH`
+The make command above will put `cu` in your `$PATH`. In order to use it, you will need to restart your terminal or run `source ~/.bashrc` (or equivalent for your shell).
```
tiborvass: I would do `hash -r` instead of `source`. It works at least on zsh and bash. I'm pretty sure people using other shells can figure it out :)

jasonmccallister: Nice, I did not know that was a thing. Adding it now!

aluzzardi:
Thank you 🙏

README.md
```diff
@@ -40,7 +40,13 @@ It's an open-source MCP server that works as a CLI tool with Claude Code, Cursor
 make
 ```
 
-Make sure to put `cu` in your `$PATH`
+This will build the `cu` binary but _NOT_ install it to your `$PATH`. If you want to build and install the binary into your `$PATH`, run:
+
+```sh
+make install
```
tiborvass: `make install && hash -r` ? People usually don't read past the code snippets.

tiborvass:
```suggestion
make install && hash -r
```

PR #15, opened 2025-06-06
fix: goose configuration

eunomie:
With `envs: []` I have the following error:

    $ goose

    thread 'main' panicked at /Users/runner/work/goose/goose/crates/goose-cli/src/session/builder.rs:147:14:
    should load extensions: Failed to deserialize value: invalid type: sequence, expected struct Envs
    note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

With `envs: {}` this is working fine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ extensions:
     cmd: cu
     args:
     - stdio
-    envs: []
+    envs: {}
 ```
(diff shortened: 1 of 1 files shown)


shykes:
Thank you!

Issue #16, opened 2025-06-06
`watch` command not found


eunomie:
When running `cu watch`:

```
$ cu watch
Error: exec: "watch": executable file not found in $PATH
Usage:
  cu watch [flags]
```

This is because the `watch` binary is not available on the machine by default.

I've fixed it by a simple `brew install watch` but that's Mac only.

I'm not sure what's the best approach, should we warn during `make build` that `watch` is required? Or to use an alternative in Go (if that exists) so it can be built at the same time?

jasonmccallister:
Might be worth checking if the watch program is installed and prompting the user to install if not, made a quick PR to make the error more friendly.

aluzzardi:
All `cu watch` does is wrap a `git log` command around a `watch -n1` -- should be very trivial to get rid of the `watch` dependency and replace that with a for loop + `time.Sleep(1*time.Second)`

cwlbraa:
yeah, the `watch` dependency is a hack here. my initial impl did, in fact, use a ticker and a for loop but it had render problems (flashing) that I didn't have time to debug. the way to fix this is to remove the watch dependency and do the rendering properly.

jasonmccallister:
Ok, I can take a pass at that. Debugging some Claude issues with `cu` this morning before I can/could test the watch functionality.

tiborvass:
Fixed in #17

PR #17, opened 2025-06-06
Do not rely on a watch binary to exist

jasonmccallister:
Return a helpful error if the `watch` command is not installed on the users machine.

Helps with #16.


tiborvass:
@jasonmccallister thanks! Probably easier to just implement it in a Go for loop.

tiborvass:
I implemented `watch` in Go at https://github.com/tiborvass/go-watch so we can use it here.

This would fix #16

PR #33, opened 2025-06-07
Implement a delete sub-command for deleting environments.

brendandburns:
I can also add this to the MCP server, not sure if that is desirable or not.

(also, this project probably needs some unit tests :)

### DIFF
--- a/environment/git.go
+++ b/environment/git.go
@@ -35,6 +35,45 @@ func (env *Environment) GetWorktreePath() (string, error) {
 	return homedir.Expand(fmt.Sprintf("~/.config/container-use/worktrees/%s", env.ID))
 }
 
+func (env *Environment) DeleteWorktree() error {
+	worktreePath, err := env.GetWorktreePath()
+	if err != nil {
+		return err
+	}
+	parentDir := filepath.Dir(worktreePath)
+	fmt.Printf("Deleting parent directory of worktree at %s\n", parentDir)
+	return os.RemoveAll(parentDir)
+}
+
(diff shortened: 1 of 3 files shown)


environment/git.go
```diff
@@ -35,6 +35,45 @@ func (env *Environment) GetWorktreePath() (string, error) {
...
 }
 
+func (env *Environment) DeleteWorktree() error {
+	worktreePath, err := env.GetWorktreePath()
+	if err != nil {
+		return err
+	}
+	parentDir := filepath.Dir(worktreePath)
+	fmt.Printf("Deleting parent directory of worktree at %s\n", parentDir)
+	return os.RemoveAll(parentDir)
```
cwlbraa, line 45: this is the only thing surprising to me about this impl, using `rm` instead of `git worktree remove` -- that said it's pretty [clearly documented](https://git-scm.com/docs/git-worktree#_description) that this approach is safe and prune will clean up the admin files in cuRepoPath.

cwlbraa (approved):

cwlbraa:
> (also, this project probably needs some unit tests :)

couldn't agree more lol

environment/git.go
brendandburns: Happy to update this to use `git worktree remove` if you want. This was based on my own experimentation with what was the easiest thing to do, but happy to update.

cwlbraa: ultimately having played with it, i find your approach to be preferable just because we don't have typed errors coming out of git commands, so os.RemoveAll followed by prune makes it much easier to keep things re-entrant.

PR #38, opened 2025-06-08
cmd/cu: improve signal handling to recover from raw term

tiborvass:
Follow up to #17

### DIFF
--- a/cmd/cu/main.go
+++ b/cmd/cu/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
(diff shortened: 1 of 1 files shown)


tiborvass:
@cwlbraa correct. And for posterity too, SIGUSR1 and SIGWINCH are handled separately. SIGUSR1 in main() just before and SIGWINCH currently is handled by go-watch. I could/should refactor it to allow passing a channel instead so that signal handling responsibility stays within main() and not in the library.

PR #40, opened 2025-06-08
fix: Resolve "Raw mode is not supported" error with piped input

buddypia:
#### Overview

This PR fixes the `Raw mode is not supported` error that occurs when piping content to standard input (`stdin`). 

This allows commands like cat file.md | claude to execute correctly without needing the --dangerously-skip-permissions flag as a workaround.

#### Context

In the latest claude code version (`1.0.17`), running the command with a pipe to feed content from stdin results in an error.

**Steps to Reproduce:**
```bash
cat ./examples/hello_world.md | claude
```

**Error Output:**
```
Error: Raw mode is not supported on the current process.stdin, which Ink uses as input stream by default.
Read about how to prevent this error on https://github.com/vadimdemedes/ink/#israwmodesupported
```
Currently, this error can be bypassed by using the --dangerously-skip-permissions flag:

**Workaround:**
```bash
cat ./examples/hello_world.md | claude --dangerously-skip-permissions
```

#### Related Issue
https://github.com/anthropics/claude-code/issues/404#issuecomment-2791419263

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ curl --create-dirs -o .github/copilot-instructions.md https://raw.githubusercont
 ### Run with [Claude Code](https://www.anthropic.com/claude-code)
 
 ```console
-cat ./examples/hello_world.md | claude
+cat ./examples/hello_world.md | claude --dangerously-skip-permissions
 ```
 
 ### Run with [goose](https://block.github.io/goose/)


aluzzardi (approved):

johnrees:
a note for anyone who finds this PR and runs`cat ./examples/hello_world.md | claude --dangerously-skip-permissions` but still gets the error -

run `claude --dangerously-skip-permissions` on its own and accept the permissions prompt

Screenshot 2025-06-21 at 12 42 39

it should work after that

Issue #50, opened 2025-06-10
Changes exist in the environment but not committed to git


samalba:
I used claude+cu to implement a small feature in python. It properly created the environment and use the mcp container-use to write the changes.

If I spin up a terminal using `cu terminal <env>`, I can properly see the changes. However if I checkout the branch `git checkout <env name>`, I don't see the changes.

Looking at the claude log, it seems all the tools it called on `container-use:*` worked properly.

samalba:
Was able to debug this with the help of @aluzzardi. Basically I had one file uncomitted (CLAUDE.md that I created when setting up the project) and I was also confused by the environment's branch that is unsync'ed by default.

Not a bug, but confusing.

After I stashed my local (uncomitted) CLAUDE.md file:

```sh
git checkout <env branch>
git pull
```

Now I could see the diff and merge `cu merge <env name>`.

The confusion comes from the fact that the local branch created by `cu` isn't sync'ed with the fork. The work is being done in a fork that lives in `~/.config/container-use/repos/`.

cwlbraa:
there's a big tradeoff with syncing state to the user repo's branch: doing so might result in a simpler UX for viewing agent changes, but would immediately introduce problems if the user modifies their checkout of that branch in any way. the server would either have to:

a. violently overwrite user repo state to complete the sync.
b. best-effort sync only, meaning once you've got an uncommitted conflict, syncing stops. the server could surface a warning to the user, but only via tool responses (meaning we'd be reliant on the LLM to understand and present the warning)

the current approach dodges this by requiring that the user manage the sync via pull. they can still view "active" changes by looking directly at the remote `git log container-use/<env branch>`, which is naturally a detached head. the fact it's a detached head prevents sync errors even if the user checks it out and introduces conflicts.

cwlbraa:
i also think there's a partial solution here via ordering... you've got 1 uncommitted file, and that gets auto-committed when we create the environment, but i think that autocommit is happening _before_ the initial branch is checked out... 

would we rather have it error on `checkout` (as it does currently) or on `pull`?

samalba:
> the current approach dodges this by requiring that the user manage the sync via pull. they can still view "active" changes by looking directly at the remote git log container-use/<env branch>, which is naturally a detached head. the fact it's a detached head prevents sync errors even if the user checks it out and introduces conflicts.

[...]

PR #52, opened 2025-06-10
cmd/cu: adds show command to help with git show <remote branch>

samalba:
Related to issue #50.

I want to ignore that there is a remote-local fork under `~/.config`. So I should be able to show the local env, without having to know where it lives on the disk or understand there are 2 forks (mine and cu's).

### DIFF
--- a/cmd/cu/log.go
+++ b/cmd/cu/log.go
@@ -1,56 +1,30 @@
 package main
 
 import (
-	"fmt"
-	"io"
-	"log/slog"
 	"os"
-	"time"
+	"os/exec"
+	"strings"
+
+	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


cwlbraa:
you don't necessarily need to know that there's a remote-local fork at any place on disk in order to know that there's a remote... git is constantly going to tell you that there's commits upstream. 

also would this be better as `cu log`? show just gives you HEAD, which is a bit less useful...

samalba:
I guess cu log would do it yeah. I can make the change if others think it's necessary too.

samalba:
Finally I renamed to using git log instead.

cwlbraa (approved):

PR #53, opened 2025-06-10
set up optional default shell configuration, defaulting to bash

### LINKED ISSUE #49: "cu terminal <env>" should default to bash
Spawning a term with `cu terminal` is barely usable (no termcaps) with the default dash, until to run `bash`. We should default to `bash` instead.

cwlbraa:
fixes #49

ubuntu's default dash->sh has very poor interactive features, so while this is a little heavy-handed, it gives a much better default experience to `cu terminal` while also giving users the ability to override via prompting or configuration.

i'm scared that environment.json is a slippery slope. how many bajillion fields are we gonna add to it? how annoying is it gonna get to maintain backwards compatibility?

### DIFF
--- a/environment/environment.go
+++ b/environment/environment.go
@@ -27,6 +27,7 @@ const (
 	instructionsFile = "AGENT.md"
 	environmentFile  = "environment.json"
 	lockFile         = "lock"
+	defaultShell     = "bash"
 )
 
 type Version int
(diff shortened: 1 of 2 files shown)


samalba:
Was about to open a PR with this commit: https://github.com/dagger/container-use/commit/bb94a8acf7db288d68694fe0394d52dab16d5906

Do we need the extra field in the env config?

cwlbraa:
@samalba 

"need" is a strong word, but the extra field gives folks configurability to use zsh or fish and bring in their own dotfiles. your approach is much simpler, but is basically "force bash if available". 

i don't particularly like adding things to the env json, though. 🤷 

also fyi that .bash_aliases line right above the diff is broken, it doesn't seem to set the PS1 for my bash at least. I think it's supposed to be writing .bashrc?

tiborvass:
Since #67 is merged, I'll close this for now. Feel free to reopen it if we think we need configurability.

PR #55, opened 2025-06-11
cmd/cu: fixes merge misleading error and exit code

### LINKED ISSUE #28: cu merge causes error with exit status 1
After doing `make install && hash -r` from latest source, I tried a very simple example with claude that ended successfully with:

```bash
⏺ Done! I've created a simple Python hello world script. To view your work:

  git checkout python-hello-world/saving-racer

  The script hello.py prints "Hello, World!" when run.
```

Then from a different terminal window opened into the same dir, I tried `cu merge python-hello-world/saving-racer` and got this error:

```
> cu merge python-hello-world/saving-racer

Error: exit status 1
Usage:
  cu merge <env> [flags]

Flags:
  -h, --help   help for merge

exit status 1
```

I wonder if I'm doing something wrong?

samalba:
Fixes #28 

When there is nothing to merge, no message is written on stdout (it took me some time to realize the changes were already merged).

And when there is nothing to stash, the `stash pop` exit 1, which displays the help for the merge command.

This is misleading and confusing, this fixes it (prints stdout and always exit 0 on `stash pop`, behaves like `stash`).

### DIFF
--- a/cmd/cu/merge.go
+++ b/cmd/cu/merge.go
@@ -17,9 +17,10 @@ var mergeCmd = &cobra.Command{
 		env := args[0]
 		// prevent accidental single quotes to mess up command
 		env = strings.Trim(env, "'")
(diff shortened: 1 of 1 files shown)


tiborvass (approved):
I pushed a commit to remove bash wrapper.

PR #62, opened 2025-06-12
environment management overhaul

aluzzardi:
This overhauls how state and environments are managed.

- Environments are "resumed" from state each and every time, rather than keeping the state hanging in memory. This makes resuming safe (whether it's the same session later on, a different agent, etc)
- `Repository` contains CRUD operations around environments
- All the git logic is stashed in there. Environment can only operate on a given worktree


repository/util.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// Otherwise, let's use the normalized origin as path
+	normalizedOrigin, err := normalizeGitURL(strings.TrimSpace(origin))
+	if err != nil {
+		return "", err
+	}
+	return homedir.Expand(filepath.Join(cuRepoPath, normalizedOrigin))
+}
+
+func normalizeGitURL(endpoint string) (string, error) {
```
aluzzardi: This boilerplate was copied from dagger, which in turn was copied from Go.

It simply normalizes e.g. `https://github.com/dagger/container-use.git` and `git@github.com:dagger/container-use.git` both to `github.com/dagger/container-use`

environment/environment.go
```diff
@@ -67,64 +66,64 @@ type Environment struct {
...
-		container:   newState,
-	}
-	containerID, err := revision.container.ID(ctx)
+func (n *Notes) Add(format string, a ...any) {
+	*n = append(*n, fmt.Sprintf(format, a...))
+}
+
+func (n *Notes) Clear() {
+	*n = []string{}
+}
```
cwlbraa: don't these need mutex protection?

environment/environment.go
```diff
@@ -67,64 +66,64 @@ type Environment struct {
...
+
+func (n *Notes) String() string {
+	return strings.Join(*n, "\n")
+}
+
+func (env *Environment) Export(ctx context.Context) (rerr error) {
+	_, err := env.container.Directory(env.Config.Workdir).Export(
+		ctx,
+		env.Worktree,
```
cwlbraa: this is a bit of a hole in the abstraction. not a problem for now but a good thing to noodle on.

repository/repository.go
```diff
@@ -0,0 +1,187 @@
...
+		return nil, err
+	}
+
+	envs := []string{}
+	for _, branch := range strings.Split(branches, "\n") {
+		branch = strings.TrimSpace(branch)
+		// FIXME(aluzzardi): This logic is broken
+		if !strings.Contains(branch, "/") {
+			continue
+		}
```
cwlbraa, line 166: broken because non-env-name branches with slashes are gonna trip things up?

mcpserver/tools.go
```diff
@@ -516,194 +506,204 @@ Background commands are unaffected by filesystem and any other kind of changes.
 				string(out), env.Config.Workdir, env.ID)), nil
 		}
 
+		// We want to update the repository whether the command fails or not
+		defer func() {
+			_ = repo.Update(ctx, env, "Run "+command, request.GetString("explanation", ""))
```
cwlbraa: 🤔 i was expecting to see a lot more of this call given we've killed the propogateToXYZ method... is there only a few of them just because you've commented a lot of things out?

mcpserver/tools.go
```diff
@@ -479,19 +467,21 @@ Failure to do so will result in the tool being stuck, awaiting for the command t
...
 		}
+
 		command := request.GetString("command", "")
 		shell := request.GetString("shell", "sh")
 
 		background := request.GetBool("background", false)
 		if background {
+			// We want to update the repository whether the command fails or not
+			defer func() {
+				_ = repo.Update(ctx, env, "Run "+command, request.GetString("explanation", ""))
```
cwlbraa: update calls definitely need error handling unless i'm missing something...

environment/environment.go
aluzzardi: Done

cwlbraa (approved):
directionally this looks right to me, accomplishes the more immediate goals of #59 without trying to dot every i and cross every t. 

imo we need to get this and #65 onto main ASAP to clear a path for more incremental work... eg i wanna get the "real" #11 environment resume working. with this PR plus the tests, the impl will be much cleaner (although it'll still involve similar dirty tricks)

parts of my Remote refactor will fit in here as a sub-package of Repository, i think.

mcpserver/tools.go
aluzzardi: Just pushed a change to remove dead code. AFAIK, we only have `run`, `file_write`, `file_delete`, `env_update` and `service_add` which are covered here.

env creation is covered by `repo.Create`, and everything else is read-only.

mcpserver/tools.go
aluzzardi: Workaround to make git notes propagate on command failure, e.g. calling it whether the command works or not.

I agree this is too naive

repository/repository.go
aluzzardi: I think this is a comment from before the refactoring ... not sure why it was broken in the first place :D

I think because it assumes a specific naming schema, and also branches with slashes will break it.

environment/environment.go
aluzzardi: Agreed.

I was thinking it's not *that* bad because the interface of environments is pretty straightforward: **worktree in** (`New`, `Load`), **worktree out** (`Export`). The rest is out of scope for the environment.

cwlbraa: the hard part, and a big part of why #59 felt like boiling the ocean, is eventually it'd great to point these at a network remote and these lil abstraction violations will prevent us from doing that.

mcpserver/tools.go
cwlbraa: we at least gotta be logging errors from this, if not populating rerr with some sort of multi-wrapped error on Update failure?

environment/environment.go
aluzzardi: Yeah. I actually kept #59 in mind while doing this -- I was kinda hoping that this would be compatible. 1) Keep the environment dumb (load stuff <here>, dump stuff <there>). 2) All that logic moved as-is (for now) in a layer above 3) Eventually `Repository` could use the `Remote` abstraction (e.g. rather than integrating `Remote` in all of `Environment`, we could integrate it into `Repository` only?)

Just realized that maybe I'm missing something. I assumed we'd still use local storage as a staging area, even for other storage implementation, but I see now you meant to be able to avoid doing that and going straight to the alternative storage without stopping by local disk?

mcpserver/tools.go
aluzzardi: Fixed -- PTAL

mcpserver/tools.go
```diff
@@ -164,15 +183,39 @@ func EnvironmentToCallResult(env *environment.Environment) (*mcp.CallToolResult,
...
 The environment is the result of a the setups commands on top of the base image.
 Read carefully the instructions to understand the environment.
 DO NOT manually install toolchains inside the environment, instead explicitly call environment_update`,
 		),
 		mcp.WithString("explanation",
-			mcp.Description("One sentence explanation for why this environment is being opened or created."),
+			mcp.Description("One sentence explanation for why this environment is being created."),
 		),
-		mcp.WithString("source",
+		mcp.WithString("environment_source",
```
cwlbraa, line 218: one day we're gonna have evals so we can make these changes on more than just vibes XD

mcpserver/tools.go
```diff
@@ -498,8 +435,12 @@ Failure to do so will result in the tool being stuck, awaiting for the command t
 					ports = append(ports, int(port.(float64)))
 				}
 			}
-			endpoints, err := env.RunBackground(ctx, request.GetString("explanation", ""), command, shell, ports, request.GetBool("use_entrypoint", false))
-			if err != nil {
+			endpoints, runErr := env.RunBackground(ctx, request.GetString("explanation", ""), command, shell, ports, request.GetBool("use_entrypoint", false))
+			// We want to update the repository even if the command failed.
+			if resp, err := updateRepo(); err != nil {
+				return resp, nil
+			}
```
cwlbraa, line 442: noting here that in the case where we fail to update the repo, we discard runErr. 

i think that's fine for now, in theory the "failed to update repo" is the more violent and important error. it may be hard to debug without logs from environment.Run, but that failure mode is also mostly theoretical for right now.

same thing exists below for the !background case.

environment/environment.go
cwlbraa: >  I assumed we'd still use local storage as a staging area, even for other storage implementation, but I see now you meant to be able to avoid doing that and going straight to the alternative storage without stopping by local disk?

yep, that's the eventual goal. honestly it's very cart-before-horse to be trying to break your assumption right now, but that's where I started with #59. We'll get there eventually, but we also need to upstream some Git apis into dagger first anyways.

mcpserver/tools.go
aluzzardi: Yeah went back and forth between the 2. My thinking was that a failed command "happens", it's just the model iterating. However, failing to update the repo is an urgent error that needs to be brought forward to the user (something is very wrong)

PR #65, opened 2025-06-13
Add comprehensive test suite for file persistence and change tracking

grouville:
# Summary

Comprehensive test suite to ensure reliable file persistence, change tracking, and environment isolation in container-use.

### Core Functionality Tests
- **Work Persistence**: Files and changes persist across container restarts
- **Change Tracking**: Automatic capture and auditing of all modifications
- **File Handling**: Graceful handling of Python caches, binaries, and large files
- **Environment Isolation**: Safe parallel operations with proper isolation
- **Configuration**: Reliable persistence of base images and setup commands

### Test Coverage
- Git operations (error handling, worktrees, empty directories)
- Selective staging for Python cache and binary files
- Configuration persistence and isolation verification
- Behavior-driven tests focused on user experience
- Test helpers for isolated environment setup

### Implementation Notes
- Modified `git.go` to support test isolation via `CONTAINER_USE_CONFIG_DIR`
- Documented known limitations with skipped tests
- Added to GitHub workflows for CI/CD

## Status
Ready for review

## UPDATE 06/25/2025

Rewrote the entire test suite, following the new paradigm. The idea is to have integration tests that mimick what a `user` will do to end up in a bug / weird situation

Already ready for review

Introduced the `Repository: basePath`: 
```shell
type Repository struct {
	userRepoPath string
	forkRepoPath string
	basePath     string // defaults to ~/.config/container-use if empty
}
```
instead of the `CONTAINER_USE_CONFIG_DIR` to isolate tests and run them in parallel 

The `WithRepository` is a helper function that spins up Dagger -- a repo with some configurable setup function and returns a userActions, which has access to the same function as the agent would. The idea is for this abstraction to also be able to mimick the CLI calls in the near future to declaratively express any user scenario

TODO: 
- [ ] do not copy the mcp handler
- [ ] cleanup a bit


environment/test_helpers.go
```diff
@@ -0,0 +1,314 @@
+package environment
```
cwlbraa: does this break down if we put it in `package environment_test`?

environment/unit_test.go
```diff
@@ -0,0 +1,210 @@
+package environment
```
cwlbraa: the unit tests will conflict with #59, but not in a way that's hard to fix i don't think, just gotta move some shit around

environment/README.md
```diff
@@ -71,3 +71,48 @@ So the flow is: **Branch** (the logical environment) → **Worktree** (filesyste
 - `environment.go` - Core environment management
 - `git.go` - Worktree and Git integration
 - `filesystem.go` - File operations within containers
+
+## Testing
+
+The environment package includes comprehensive unit and integration tests that verify critical user behaviors:
+
+### Running Tests
```
cwlbraa: i think it's worth immediately wiring these up to CI/dagger prior to merging

environment/README.md
```diff
@@ -71,3 +71,66 @@ So the flow is: **Branch** (the logical environment) → **Worktree** (filesyste
...
+- Automatic change tracking
+- Environment isolation
+- Edge case handling (Python cache, binary files, large files)
+- Performance with 100+ file projects
+- Configuration persistence
+
+### Known Issues
+
+The test suite documents several known bugs with skipped tests:
+- Python `__pycache__` directories cause git operations to fail
```
aluzzardi: I'd either file these as issues or not list them at all, we'll forget to keep this README up to date :)

environment/README.md
```diff
@@ -71,3 +71,66 @@ So the flow is: **Branch** (the logical environment) → **Worktree** (filesyste
 - `environment.go` - Core environment management
 - `git.go` - Worktree and Git integration
 - `filesystem.go` - File operations within containers
+
+## Testing
```
aluzzardi: nit: Maybe move these to `CONTRIBUTING.md`?

environment/unit_test.go
aluzzardi: nit: Go convention, unit tests should be `environment_test.go` if we're testing the environment file, not `unit_test.go`

environment/integration_test.go
```diff
@@ -0,0 +1,729 @@
+package environment
```
aluzzardi: nit: MAYBE we should move these into their own package, e.g. `environment/integration` or something, keep the unit tests top-level. MAYBE. Thoughts?

aluzzardi (commented):
Thank you @grouville, much appreciated! Some minor comments

environment/integration_test.go
cwlbraa: imo for small projects the thing he's doing with -short is quite nice and provides a similar test grouping UX. i'd like to try it out and see if there are other drawbacks that aren't yet obvious to me... the one i can see from here is it might be too easy to share helpers that turn unit tests into int tests.

grouville: > nit: MAYBE we should move these into their own package, e.g. `environment/integration` or something, keep the unit tests top-level. MAYBE. Thoughts?

Regarding the integration tests: I’ve kept them in the environment package because they need access to the private functions to thoroughly verify the implementation. Moving them to a separate package (like environment_test or environment/integration) would limit testing to the public API only, reducing coverage (or requiring to turn more of those functions public. What would be your recommended approach (I don't mind) ?

environment/README.md
grouville: done ✅

environment/test_helpers.go
grouville: quite a lot -- as some functions i am using are not public. Happy to move them too https://github.com/dagger/container-use/pull/65#discussion_r2148877778

environment/README.md
grouville: ✅

environment/README.md
grouville: ✅

environment/unit_test.go
grouville: ✅

CONTRIBUTING.md
```diff
@@ -0,0 +1,78 @@
...
+   git clone git@github.com:dagger/container-use.git
+   ```
+3. **Install Dependencies**:
+
+   ```bash
+   go mod download
+   ```
+4. **Container Runtime**: Ensure you have a compatible container runtime installed (e.g., Docker).
+
+## Testing
```
cwlbraa, line 38: we should document building as well

```suggestion
## Building

```bash
go build ./cmd/cu`
```

## Testing
```

Comment (cwlbraa) [reply CONTRIBUTING.md:38]: lmao this is an impossible suggestion to make because the markdowns conflict, but you get the idea

File: .github/workflows/cu.yml [hunk 8]
```diff
@@ -1,4 +1,4 @@
-name: Build
+name: Build and Test
```
cwlbraa: can we have 2 workflow files so they show as separate rows in the checks panels? (both on commits and at the tail of the PR log)

CONTRIBUTING.md
cwlbraa: this would also imply we can delete the build instructions from the main readme

grouville: ✅

.github/workflows/cu.yml
grouville: ✅

cwlbraa (approved):
lgtm, just need to decide whether this or #62 goes in first... conventional wisdom says tests go in first, but idk cc @aluzzardi

repository/repository.go
```diff
@@ -25,12 +26,39 @@ const (
...
+func (r *Repository) getBasePath() string {
+	if r.basePath != "" {
+		return r.basePath
+	}
+	return cuGlobalConfigPath
+}
+
+// getRepoPath returns the path for storing repository data
+func (r *Repository) getRepoPath() string {
+	return r.getBasePath() + "/repos"
```
aluzzardi, line 43: nit: use `filepath.Join`

repository/repository.go
```diff
@@ -25,12 +26,39 @@ const (
...
+}
+
+// getWorktreePath returns the path for storing worktrees
+func (r *Repository) getWorktreePath() string {
+	return r.getBasePath() + "/worktrees"
```
aluzzardi, line 48: `filepath.Join`

repository/repository.go
```diff
@@ -25,12 +26,39 @@ const (
...
+
+// OpenWithBasePath opens a repository with a custom base path for container-use data.
+// This is useful for tests that need isolated environments.
+func OpenWithBasePath(ctx context.Context, repo string, basePath string) (*Repository, error) {
+	output, err := RunGitCommand(ctx, repo, "rev-parse", "--show-toplevel")
 	if err != nil {
-		if strings.Contains(err.Error(), "not a git repository") {
+		// Check for exit code 128 which means not a git repository
+		var exitErr *exec.ExitError
+		if errors.As(err, &exitErr) && exitErr.ExitCode() == 128 {
```
aluzzardi, line 62: nice!

aluzzardi: For **later** as a follow-up: **maybe** we should add this logic to `RunGitCommand` directly, catch all git commands that are not run from a git repository?

Although I think it's impossible since everything starts with `Open()`

repository/git.go
```diff
@@ -482,13 +485,16 @@ func (r *Repository) isBinaryFile(worktreePath, fileName string) bool {
...
+	origin, err := RunGitCommand(ctx, repo, "remote", "get-url", "origin")
 	if err != nil {
 		// If not -- this repository is a local one, we're going to use the filesystem path for the container-use repo
-		if strings.Contains(err.Error(), "No such remote") {
-			return homedir.Expand(filepath.Join(cuRepoPath, repo))
+		var exitErr *exec.ExitError
+		if errors.As(err, &exitErr) && exitErr.ExitCode() == 2 {
+			// Exit code 2 means the remote doesn't exist
+			// Use just the base name for temporary directories to avoid deeply nested paths
+			return homedir.Expand(filepath.Join(r.getRepoPath(), filepath.Base(repo)))
```
aluzzardi: So actually that was done on purpose -- we used to use the directory name, but that failed if multiple repositories had the same name on different directories (e.g. `~/work/foo` and `~/Documents/foo`).

Was there an issue this is fixing, or was it for aesthetics?

repository/repository.go
```diff
@@ -25,12 +26,39 @@ const (
 type Repository struct {
 	userRepoPath string
 	forkRepoPath string
+	basePath     string // defaults to ~/.config/container-use if empty
+}
+
+// getBasePath returns the base path for container-use data, using the default if not set
+func (r *Repository) getBasePath() string {
```
aluzzardi, line 34: nit: This logic is already handled by `Open()` (e.g. setting `cuGlobalConfigPath` when calling `Open()`).

I think it would be simpler to avoid duplication of this logic, always assume `r.basePath` is set, and get rid of `getBasePath` altogether

environment/integration/repository_test.go
```diff
@@ -0,0 +1,128 @@
+package integration
```
aluzzardi, line 1: Shouldn't these go into `repository/repository_test.go`? We're testing `Open` over there and the rest here here

cwlbraa:
> To me integration tests would be like ... testing cu with claude for instance.

i'd prolly call tests involving claude e2e, but if we're integrating anything here it's probably the MCP server against repository+env

PR #78, opened 2025-06-15
terminal: improve bash PS1 logic

tiborvass:
PS1 is overwritten by bashrc files.
This commit specifies a cu-specific bashrc file instead which first sources ~/.bashrc before setting PS1.

### DIFF
--- a/environment/environment.go
+++ b/environment/environment.go
@@ -468,15 +468,30 @@ func (env *Environment) Fork(ctx context.Context, explanation, name string, vers
 
 func (env *Environment) Terminal(ctx context.Context) error {
 	container := env.container
-	// In case there's bash in the container, show the same pretty PS1 as for the default /bin/sh terminal in dagger
-	container = container.WithNewFile("/root/.bash_aliases", `export PS1="\033[33mdagger\033[0m \033[02m\$(pwd | sed \"s|^\$HOME|~|\")\033[0m \$ "`+"\n")
-	defaultShell := []string{}
-	// Check if bash is available
-	if _, err := container.WithExec([]string{"grep", "/bash", "/etc/shells"}).Sync(ctx); err == nil {
-		defaultShell = []string{"bash"}
+	var cmd []string
+	var sourceRC string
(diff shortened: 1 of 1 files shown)


tiborvass:
I also changed `dagger` to `cu` in the PS1 for both `bash` and `sh`, and I added `bash --version` output when bash is detected, as a way to indicate that bash was selected. Maybe we just modify PS1 to include the shell but I'm not good in PS1 design.

environment/environment.go
```diff
@@ -468,15 +468,30 @@ func (env *Environment) Fork(ctx context.Context, explanation, name string, vers
...
+				cmd = []string{shell, "--rcfile", "/cu/rc.sh", "-i"}
+				break
+			}
+		}
+	}
+	// Try to show the same pretty PS1 as for the default /bin/sh terminal in dagger
+	container = container.WithNewFile("/cu/rc.sh", sourceRC+`export PS1="\033[33mcu\033[0m \033[02m\$(pwd | sed \"s|^\$HOME|~|\")\033[0m \$ "`+"\n")
+	if cmd == nil {
+		// If bash not available, assume POSIX shell
+		container = container.WithEnvVariable("ENV", "/cu/rc.sh")
```
cwlbraa, line 490: ENV layers in the file at the path using `source`? or does it replace something else?

tiborvass: Yes, ENV is a POSIX shell way of setting the rcfile.

cwlbraa: lemme rephrase: does the shell still load `~/.bashrc` when you've overwritten ENV? what about `/etc/profile`?

tiborvass: `/etc/profile` is for login shells only, and that's usually not loaded in container settings.

`~/.bashrc` is loaded explicitly in the bash case (see `sourceRC`) except PS1 is overridden AFTER that.

POSIX shells like ash or dash do not have a default rcfile for non login sessions, they only source ENV if present.

cwlbraa (approved):

PR #79, opened 2025-06-16
updating goose docs

michaelneale:
ads a deep link and also how to add it via cli


README.md
```diff
@@ -127,6 +128,20 @@ extensions:
...
+
+For the Goose desktop, paste this into your browser: 
+<code>
+goose://extension?cmd=cu&arg=stdio&id=container-use&name=container%20use&description=use%20containers%20with%20dagger%20and%20git%20for%20isolated%20environments"
+</code>
+
+Then append to your .goosehints file: 
+
+```sh
+curl -fsSL https://raw.githubusercontent.com/dagger/container-use/main/rules/agent.md >> .goosehints
```
cwlbraa: in goose, i think it already gets this prompt via the mcp init message `instructions` field. this is a big part of why, as mcp server developers, we really like using goose as our development client.

rules/agent.md
```diff
@@ -3,3 +3,5 @@ ALWAYS use ONLY Environments for ANY and ALL file, code, or shell operations—N
 DO NOT install or use the git cli with the environment_run_cmd tool. All environment tools will handle git operations for you. Changing ".git" yourself will compromise the integrity of your environment.
 
 You MUST inform the user how to view your work using `git checkout <branch_name>`. Failure to do this will make your work inaccessible to others.
+
+When starting a new session, look for git branches and run environment_list to orient yourself if not clear
```
cwlbraa: what's the intent of this prompt addition? what are you trying to get the agent to do?

michaelneale: without this - it seemed to not know it was meant to use cu - and would look around, see a (in this case) empty main branch, vs discovering environments. Not sure how automated we want this to be, but that was what I observed.

README.md
michaelneale: hrm yeah - good point, will get rid of that.

rules/agent.md
michaelneale: @cwlbraa not been an issue for you? (was mostly thinking about case where people jump between projects, end up in existing project which happens to use cu, but not using cu in other projects). If in practice that isn't an issue, could drop this too (don't usually like adding tokens if I don't have to!)

cwlbraa: I'm still not following, I never really expect the agent to go looking at branches. I do occasionally prompt "using the most recent environment ..." or "using environment task/adverb-animal ..." and then i expect the agent to resume that environment.

I'm also not following what changes when users are jumping between projects... in the fresh repo, environment_create does have some extra stuff to do behind the scenes, but basically the only difference is that there aren't any existing environments to resume, so the prompts i shared above will confuse the agent for sure, but so will "look for git branches".

I really wanna hear more about your exact application, though, there's clearly a difference of perspective going on here and I'm plagued by curse of the expert.

aluzzardi: As I mentioned, the UX of this part (how does the human and agent interact with the environments) is still pretty much up in the air, so I'm really eager to understand what you had in mind as a fresh set of eyes!

How we typically work is:

- start from the `main` branch
- ask goose to work on X
- it will create a new environment to work with
- iterate until happiness is achieved ...
- merge the work into `main`, go back to step 1

All these steps can run in parallel (multiple gooses? geese?)

We originally commented the `environment_list` tool to avoid the agent going off rails and starting to work on someone else's environment. The agent is forced to work using an `environment_id`, and there's only 2 ways to get it:

- Create a new environment
- The user prompts one

So if the user doesn't provide an environment ID, the agent is forced to create its own sandbox and work from there

aluzzardi:
LGTM @michaelneale!

Left a comment on the `environment_list` part. It's not wrong, we're just trying to understand / refine the workflow!

rules/agent.md
michaelneale: yeah - I think it is that UX part that slightly confused me - ie I did some work. Quit things, went back later and wasn't sure how to get back (as it was still on a branch). I don't know if this prompt shoudl be in here - I think if environment_list tool is available that may be enough of a hint?

aluzzardi: Oh I see! Basically: new goose session, existing environment, how to continue?

I think right now the only way is to prompt your way back into it ("<prompt> ... in the x-y-z environment").

I don't know if `environment_list` would be enough of a hint **if** there are multiple environments to choose from (e.g. how will it resume in the right one and not step into a neighbor?)

This is at the the center of what's clearly not well defined now.

> Quit things, went back later and wasn't sure how to get back (as it was still on a branch)

The tricky part is you could mean two different things: You'd like to continue the work in your previous environment OR you may want to start new work in a brand a new environment.

Since intent is ambiguous, we default into assuming "new work in a brand new environment". If the user wants to continue work in an existing environment, we rely on either 1) the user resuming the goose session, which will pick up where it left or 2) the user explicitly specifying the environment in the prompt

I realize this is confusing and really looking for a better solution.

Maybe what could work better is if container-use provided a "current environment" to the agent. The human could switch to a branch to resume the work and that's what the "current environment" would be?

michaelneale:
@aluzzardi yeah makes sense now - I reverted that change as it looks like will be upstream changes that address all that anyway, so looking good!

PR #87, opened 2025-06-16
ci: fix brew tap PR goreleaser configuration

cwlbraa:
this is my last push and pray attempt. upon closer inspection of the goreleaser docs instead of just the yaml schema, the issue here was that we were opening a PR from main onto main. 

https://goreleaser.com/customization/homebrew_casks/
```yaml
   # Make sure the 'branch' property is different from base before enabling
   # it.
   pr:
```
if this doesn't work I'll set up my fork-oriented testing again, it's just a bit of a hassle.


tiborvass (approved):
🤷‍♂️

PR #91, opened 2025-06-17
ai generated windows build, untested

cwlbraa:
i asked claude to cross compile to windows.

it hit a build error in go-watch (makes sense, a lot of unix-specific tty stuff in there)

but then it kept going and gave me this. if you've got a windows dev setup and windows knowledge you could check out this branch,

```
GOOS=windows go build -o cu.exe ./cmd/cu
```

and see what's broken?


grouville:
We could test today

cwlbraa:
@jpadams and i just gave it a quick test. one problem is `/tmp` doesn't exist by default on windows, but after creating that `cu.exe stdio` starts up and runs. 

the windows `watch` impl the agent added is not great, it flashes a lot.

cwlbraa:
https://github.com/dagger/container-use/blob/f1baa630d43d77a86c2fc79a9175904aa7676f58/environment/git.go#L513-L515 is also problematic on windows when users have untracked files. `cp` is not a thing there.

PR #93, opened 2025-06-17
document autoApprove config

kpenfound:


levlaz:
Lets pause this one because it seems like at least in Q it is not supported all the way. found this related issue: https://github.com/aws/amazon-q-developer-cli/issues/1938

PR #96, opened 2025-06-17
fix: add NoCache option to host directory uploads

grouville:
Rebased on top of https://github.com/dagger/container-use/pull/65

_> Maybee it is being fixed by the new environment_

## Summary
- Fixes caching issue where uploading modified local directories would use stale cached versions
- Adds `NoCache: true` to all `dag.Host().Directory()` calls in `urlToDirectory()`
- Enables and updates integration test that was previously skipped due to this issue

## Test plan
- [x] Enabled and fixed `TestUploadAfterModification` integration test
- [x] Test passes with the fix, fails without it (verified)
- [x] Added edge case documentation for worktree subdirectory uploads (if users think it still makes sense) -> users _might_ want to change something on the worktree, what's the expected behavior ?


cwlbraa:
the upload tools fell out in the repository refactor, so this will need to be reexamined.

that said:

i'm not entirely sure this is the right way to fix the test... NoCache definitely gives correct behavior for Upload, but i'm fairly certain that url transformer _was_ also used for environment_source? (keyword was)

if that's true, the most recent container should have all the accumulated changes from the run_command and file_write tools, right? so busting the cache and re-uploading the base directory with those changes creates a weird situation where we then go and re-apply those writes on top of a different base.

aluzzardi:
Is this still needed?

grouville:
Not needed anymore

PR #106, opened 2025-06-19
cli: sanity checks and cleanup

aluzzardi:
- Use the "git root" as the base directory
- Add sanity checks to ensure we're running from a git repo, clear error
  message if not.
- Make sure all commands go through `repository.Open()` to ensure sanity
  checks are being run.


cwlbraa (approved):
lgtm

PR #107, opened 2025-06-19
environment description

aluzzardi:
environment description
    
- `cu list` will display each environment's short description, along
  with timestamps
- Add new information to state, updated the plumbing to carry that over

```
$ cu list
ID                            DESCRIPTION                                                  CREATED         UPDATED
flask-app/awaited-puma        Creating a simple Flask application                          10 minutes ago  10 minutes ago
flask-blue-app/current-moth   A Flask application with a stunning blue design              5 minutes ago   4 minutes ago
```

DEPENDS ON #106

### DIFF
--- a/cmd/cu/list.go
+++ b/cmd/cu/list.go
@@ -36,13 +36,24 @@ var listCmd = &cobra.Command{
 
 		defer tw.Flush()
 		for _, env := range envs {
-			fmt.Fprintf(tw, "%s\t%s\t%s\t%s\n", env.Name, env.State.Title, humanize.Time(env.State.CreatedAt), humanize.Time(env.State.UpdatedAt))
+			fmt.Fprintf(tw, "%s\t%s\t%s\t%s\n", env.Name, truncate(app, env.State.Title, 40), humanize.Time(env.State.CreatedAt), humanize.Time(env.State.UpdatedAt))
 		}
 		return nil
 	},
 }
 
+func truncate(app *cobra.Command, s string, max int) string {
+	if noTrunc, _ := app.Flags().GetBool("no-trunc"); noTrunc {
+		return s
+	}
+	if len(s) > max {
+		return s[:max] + "…"
+	}
+	return s
+}
+
 func init() {
 	listCmd.Flags().BoolP("quiet", "q", false, "Display only environment IDs")
+	listCmd.Flags().BoolP("no-trunc", "", false, "Don't truncate output")
 	rootCmd.AddCommand(listCmd)
 }


mcpserver/tools.go
```diff
@@ -266,7 +276,9 @@ var EnvironmentUpdateTool = &Tool{
 			mcp.Description("The instructions for the environment. This should contain any information that might be useful to operate in the environment, such as what tools are available, what commands to use to build/test/etc"),
 			mcp.Required(),
 		),
-
+		mcp.WithString("description",
+			mcp.Description("Short description of the work that is happening in this environment.."),
+		),
 		mcp.WithString("base_image",
 			mcp.Description("Change the base image for the environment."),
 			mcp.Required(),
```
cwlbraa, line 285: "keep this description updated using environment_update" kinda implies the agent can update just the description, no?

we can punt this to a later PR, but it feels like we're _inches_ away from removing these `mcp.Required()`s, having sane defaults/don't update if empty, and having LLMs able to update the env description without updating the rest of the setup.

environment/state.go
```diff
@@ -9,7 +9,26 @@ import (
 )
 
 type State struct {
-	Container string `json:"container"`
+	Container   string    `json:"container,omitempty"`
+	Description string    `json:"description,omitempty"`
```
cwlbraa: when we went down to just "Container" i was curious if we were gonna wanna bring back other fields. these feel right to me.

environment/environment.go
```diff
@@ -129,9 +119,15 @@ func (env *Environment) apply(ctx context.Context, name, explanation, output str
...
+	containerID, err := newState.ID(ctx)
+	if err != nil {
+		return err
+	}
+
 	env.mu.Lock()
 	defer env.mu.Unlock()
-	env.container = newState
+	env.State.UpdatedAt = time.Now()
+	env.State.Container = string(containerID)
```
cwlbraa, line 130: with HostDir.NoCache, this might have weird downstream durability requirements very similar to the ones i was talking about [here](https://discord.com/channels/707636530424053791/1380291254625239060/1385035221942145034)... i'm gonna try to test it out

cwlbraa: tested both my PR and this one, in both cases the sync'd IDs continue to be restorable through cache deletion (both `dagger core engine local-cache prune` and `docker kill $engine`)

when those things happen, i do suspect that container-use is loading a different, newer HostDir into it's initial state, and so there may be ways for restore operations to fail in interesting ways, but there are ways to fix that and we'll cross that bridge when we get to it.

cwlbraa (approved):
lgtm, though i would probably just take this a lil further and get environment_update working with less required fields.

mcpserver/tools.go
```diff
@@ -215,6 +217,10 @@ DO NOT manually install toolchains inside the environment, instead explicitly ca
 		mcp.WithString("explanation",
 			mcp.Description("One sentence explanation for why this environment is being created."),
 		),
+		mcp.WithString("description",
+			mcp.Description("Short description of the work that is happening in this environment. Keep this description updated using `environment_update`."),
```
cwlbraa: if we do punt on the less-required-arguments environment_update tool, i think we should probably remove "keep this description updated using environment_update" or else some agents might reconfigure the environment too frequently.

environment/environment.go
```diff
@@ -81,20 +86,11 @@ func (env *Environment) Export(ctx context.Context) (rerr error) {
...
 }
 
-func (env *Environment) State(ctx context.Context) ([]byte, error) {
-	containerID, err := env.container.ID(ctx)
-	if err != nil {
-		return nil, err
-	}
+func (env *Environment) container() *dagger.Container {
+	env.mu.RLock()
+	defer env.mu.RUnlock()
```
cwlbraa, line 91: ah shit, i have this exact thing on a branch with a slightly different `return`... let's both try to have a light touch with callsites or else it's gonna be a very annoying merge conflict.

i've also got updateContainer() with the mu wrapping fyi

environment/state.go
```diff
@@ -9,7 +9,26 @@ import (
 )
 
 type State struct {
-	Container string `json:"container"`
+	Container string    `json:"container,omitempty"`
+	Title     string    `json:"title,omitempty"`
```
cwlbraa, line 13: this is the right place for Title (goes into notes) but it's kinda weird that it comes from environment.Update (where all the other args go into EnvironmentConfig->git)

aluzzardi: Agreed. I don't have a better solution in mind right now though, suggestions?

cwlbraa: intuitively im starting to think there should be 2 update tools, but not sure what their names should be

PR #108, opened 2025-06-19
cu checkout

aluzzardi:
- Do not set up remote tracking branches for each environment anymore
- Instead, provide an explicit `cu checkout` command that does that
- Similar to `gh pr checkout`

This still requires reworking the prompt engineering behind the scenes

DEPENDS ON #107

### DIFF
--- a/repository/repository.go
+++ b/repository/repository.go
@@ -206,3 +206,24 @@ func (r *Repository) Delete(ctx context.Context, id string) error {
 	}
 	return nil
 }
+
+func (r *Repository) Checkout(ctx context.Context, id string) (string, error) {
+	if err := r.exists(ctx, id); err != nil {
+		return "", err
+	}
+
+	branch := "cu-" + id
+
+	// set up remote tracking branch if it's not already there
+	_, err := runGitCommand(ctx, r.userRepoPath, "show-ref", "--verify", "--quiet", fmt.Sprintf("refs/heads/%s", branch))
+	if err != nil {
+		_, err = runGitCommand(ctx, r.userRepoPath, "branch", "--track", branch, fmt.Sprintf("%s/%s", containerUseRemote, id))
+		if err != nil {
+			return "", err
+		}
+	}
+
+	_, err = runGitCommand(ctx, r.userRepoPath, "checkout", branch)
+
+	return branch, err
+}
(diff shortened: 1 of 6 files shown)


repository/repository.go
```diff
@@ -203,3 +203,24 @@ func (r *Repository) Delete(ctx context.Context, id string) error {
...
+	// set up remote tracking branch if it's not already there
+	_, err := runGitCommand(ctx, r.userRepoPath, "show-ref", "--verify", "--quiet", fmt.Sprintf("refs/heads/%s", branch))
+	if err != nil {
+		_, err = runGitCommand(ctx, r.userRepoPath, "branch", "--track", branch, fmt.Sprintf("%s/%s", containerUseRemote, id))
+		if err != nil {
+			return "", err
+		}
+	}
+
+	_, err = runGitCommand(ctx, r.userRepoPath, "checkout", branch)
```
cwlbraa: i think this impl might be overcomplicating things. do we get correct behavior with just plain `git checkout $branch`? maybe with a fetch thrown in? iirc the show-ref and `branch --track` wackiness was only necessary because I wanted the branch setup entirely in the background, and i only wanted to touch it if the user hadn't already done stuff with it. in this context, the user is doing things themselves, so we can be way less sneaky if we'd like to.

also, we should make sure this command always tries to get the latest possible remote state, maybe erroring if the user has conflicts/untracked work?

repository/repository.go
```diff
@@ -203,3 +203,24 @@ func (r *Repository) Delete(ctx context.Context, id string) error {
 	}
 	return nil
 }
+
+func (r *Repository) Checkout(ctx context.Context, id string) (string, error) {
```
cwlbraa, line 210: this makes me wanna have a separate UserRepository type...

in my head the ideal Repository is centered around the remote, whereas this method is way more centered around the UserRepository.

cwlbraa:
i'm taking this

cwlbraa:
fixed up prompt engineering and added logic to `git merge --ff-only` when the remote has commits you don't have. this logic prevents the `cu checkout` command from checking out old work after `environement_open`.

aluzzardi:
LGTM

cwlbraa (approved):

PR #109, opened 2025-06-19
remove environment names

aluzzardi:
The environment name was used to easily recognize environments. Since
now they have a full blown description, it's not needed anymore.

DEPENDS ON #107

### DIFF
--- a/mcpserver/tools.go
+++ b/mcpserver/tools.go
@@ -4,10 +4,8 @@ import (
 	"context"
 	_ "embed"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"log/slog"
-	"strings"
 
 	"github.com/dagger/container-use/environment"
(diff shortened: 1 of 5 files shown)


mcpserver/tools.go
```diff
@@ -16,42 +14,6 @@ import (
 	"github.com/mark3labs/mcp-go/server"
 )
 
-func validateName(name string) error {
```
cwlbraa, line 19: 1 less LLM slop <3

cwlbraa (approved):
woo, a user-facing breaking change! i like this both as a redundancy elimination thing and because i often get confused by the environment names that agents choose. 

for example, I'll end up with 2 envs flask-app/adj-animal and flask-dev/adj-animal, but think that they should both be prefixed `flask-app`, so i think the `flask-dev` one is broken somehow.

PR #122, opened 2025-06-24
explicit client passing

cwlbraa:
previously we were using an initialize fn and package globals to pass in the dagger client. this makes the data passing explicit so that we can avoid client initialization when it's unnecessary. 

to get that done, i've introduced an EnvironmentInfo struct that's essentially all the metadata for an environment: config, state, id, and worktree. those can be fetched and marshalled without use of the dagger client. 

Services are kinda the final frontier of this PR... right now they have actual *dagger.Service on them, but ideally we should move the config into environment info and leave the hydrated `[]*dagger.Service` on environment or something like that, but the correct data modeling to get there is not entirely clear to me yet.

### DIFF
--- a/environment/environment.go
+++ b/environment/environment.go
@@ -27,7 +27,7 @@ type EnvironmentInfo struct {
 type Environment struct {
 	*EnvironmentInfo
 
-	Client *dagger.Client
+	dag *dagger.Client
 
 	Services []*Service
 	Notes    Notes
@@ -47,7 +47,7 @@ func New(ctx context.Context, client *dagger.Client, id, title, worktree string)
 				UpdatedAt: time.Now(),
 			},
 		},
-		Client: client,
+		dag: client,
 	}
 
 	if err := env.Config.Load(worktree); err != nil {
@@ -92,7 +92,7 @@ func (env *Environment) container() *dagger.Container {
 	env.mu.RLock()
 	defer env.mu.RUnlock()
 
-	return env.Client.LoadContainerFromID(dagger.ContainerID(env.State.Container))
+	return env.dag.LoadContainerFromID(dagger.ContainerID(env.State.Container))
 }
 
 func Load(ctx context.Context, client *dagger.Client, id string, state []byte, worktree string) (*Environment, error) {
@@ -103,7 +103,7 @@ func Load(ctx context.Context, client *dagger.Client, id string, state []byte, w
 			Config:   DefaultConfig(),
 			State:    &State{},
 		},
-		Client: client,
+		dag: client,
 	}
 	if err := env.Config.Load(worktree); err != nil {
 		if !errors.Is(err, os.ErrNotExist) {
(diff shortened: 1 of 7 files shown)


aluzzardi (approved):

environment/environment.go
```diff
@@ -88,27 +92,53 @@ func (env *Environment) container() *dagger.Container {
...
 	defer env.mu.RUnlock()
 
-	return dag.LoadContainerFromID(dagger.ContainerID(env.State.Container))
+	return env.dag.LoadContainerFromID(dagger.ContainerID(env.State.Container))
 }
 
-func Load(ctx context.Context, id string, state []byte, worktree string) (*Environment, error) {
+func Load(ctx context.Context, client *dagger.Client, id string, state []byte, worktree string) (*Environment, error) {
 	env := &Environment{
+		EnvironmentInfo: &EnvironmentInfo{
```
aluzzardi: nit: can probably just call `LoadInfo` and set env.dag to avoid duplication

environment/environment.go
```diff
@@ -14,36 +14,40 @@ import (
...
+	dag *dagger.Client
 
 	Services []*Service
 	Notes    Notes
 
 	mu sync.RWMutex
 }
 
-func New(ctx context.Context, id, title, worktree string) (*Environment, error) {
+func New(ctx context.Context, client *dagger.Client, id, title, worktree string) (*Environment, error) {
```
aluzzardi: nit: can probably call it `dag` throughout (or `client` throughout)

environment/environment.go
```diff
@@ -129,7 +159,7 @@ func (env *Environment) apply(ctx context.Context, name, explanation, output str
 	return nil
 }
 
-func containerWithEnvAndSecrets(container *dagger.Container, envs, secrets []string) (*dagger.Container, error) {
+func containerWithEnvAndSecrets(client *dagger.Client, container *dagger.Container, envs, secrets []string) (*dagger.Container, error) {
```
aluzzardi: nit: can probably call it `dag` throughout (or `client` throughout)

mcpserver/tools.go
```diff
@@ -47,15 +52,15 @@ type Tool struct {
 	Handler    server.ToolHandlerFunc
 }
 
-func RunStdioServer(ctx context.Context) error {
+func RunStdioServer(ctx context.Context, client *dagger.Client) error {
```
aluzzardi: nit: can probably call it `dag` throughout (or `client` throughout)

mcpserver/tools.go
```diff
@@ -70,15 +75,26 @@ func registerTool(tool ...*Tool) {
...
+				slog.Info("Tool finished", "tool", tool.Definition.Name)
 			}()
-			return t.Handler(ctx, request)
+			return tool.Handler(ctx, request)
+		},
+	}
+}
+
+// keeping this modular for now. we could move tool registration to RunStdioServer and collapse the 2 wrapTool functions.
+func wrapToolWithClient(tool *Tool, client *dagger.Client) *Tool {
```
aluzzardi: nit: can probably call it `dag` throughout (or `client` throughout)

repository/repository.go
```diff
@@ -115,7 +116,31 @@ func (r *Repository) exists(ctx context.Context, id string) error {
 	return nil
 }
 
-func (r *Repository) Get(ctx context.Context, id string) (*environment.Environment, error) {
+// Create creates a new environment with the given description and explanation.
+// Requires a dagger client for container operations during environment initialization.
+func (r *Repository) Create(ctx context.Context, client *dagger.Client, description, explanation string) (*environment.Environment, error) {
```
aluzzardi: nit: can probably call it `dag` throughout (or `client` throughout)

repository/repository.go
```diff
@@ -115,7 +116,31 @@ func (r *Repository) exists(ctx context.Context, id string) error {
...
+		return nil, err
+	}
+
+	return env, nil
+}
+
+// Get retrieves a full Environment with dagger client embedded for container operations.
+// Use this when you need to perform container operations like running commands, terminals, etc.
+// For basic metadata access without container operations, use Info() instead.
+func (r *Repository) Get(ctx context.Context, client *dagger.Client, id string) (*environment.Environment, error) {
```
aluzzardi: nit: can probably call it `dag` throughout (or `client` throughout)

PR #128, opened 2025-06-24
environment: persistent run command changes

aluzzardi:
@cwlbraa I know you're reworking this part so feel free to discard -- I had to make this change locally otherwise I couldn't dogfood

### DIFF
--- a/environment/environment.go
+++ b/environment/environment.go
@@ -236,6 +236,9 @@ func (env *Environment) Run(ctx context.Context, explanation, command, shell str
 		}
 		return "", err
 	}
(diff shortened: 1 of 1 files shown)


cwlbraa:
after this change, are you also having issues around what gets returned when commands fail?

i've been using this test script `"[[ -f .ran ]] && exit 1 || touch .ran"` and the exit 1 seems to not make the tool calls return error text, but i can't tell how much of that is my changes,  this specific test case, or a larger problem on main.

cwlbraa:
this will be subsumed by #129

PR #130, opened 2025-06-24
cu: environment shell completion

aluzzardi:


cwlbraa:
im gonna slap an extra commit on here that moves the completion installs into install.sh + the brew tap

cwlbraa:
or wait... maybe i should just open that as a separate PR... it feels like one of those things where we'd need to merge it right before cutting a release

cwlbraa:
also fyi this doesn't work with my fzf setup, but i think that's a problem with my fzf setup more than it's a problem with this PR

cwlbraa (approved):

PR #133, opened 2025-06-25
shell completion installs

cwlbraa:
requires #130 

this PR installs shell completions using homebrew post install hooks and provides instructions on how to install them as install.sh output.

complexities:
- bash: there are canonical locations for bash, but those locations require the (very standard) [bash-completion](https://github.com/scop/bash-completion#q-where-should-i-install-my-own-local-completions) plugin.
   - in homebrew, we'll just assume you have that package or if you're using bash over default zsh, you'll guess that you need it.
- zsh: there are no canonical locations, but you can look up locations from fpath if your shell is all configured (ohmyzsh, antigen, etc). 
   - homebrew [documents](https://docs.brew.sh/Shell-Completion#configuring-completions-in-zsh) how things need to be set up for zsh completions, and so long as the user has `eval "$(brew shellenv)"` in their zshrc, we can write to `$(brew --prefix)/share/zsh/site-functions` pretty blindly
- fish is friendly so it's got canonical paths 🎉 

because it's complex to detect what zsh/bash setups a given installer has from inside the installer script, we instead have the install script tell users both the "dumb" rc-centric way to install, and the most common "managed" way.

one alternative approach that folks have talked about is putting this in a `cu init` subcommand or similar. we could do that, but it has all the same difficulties that come with putting this in autoinstalling from our bash-specific install script. there are too many possible shells, userspace package managers, and plugin frameworks and custom configurations (including the absence of them).

NOTE: i haven't tested the homebrew tap, don't merge this without testing it on a fork

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -6,6 +6,7 @@ project_name: container-use
 before:
   hooks:
     - go mod tidy
+    - ./scripts/completions.sh
 
 builds:
   - id: cu
@@ -34,6 +35,7 @@ archives:
     files:
       - README.md
       - LICENSE
+      - completions/*
 
 homebrew_casks:
   - repository:
@@ -53,9 +55,14 @@ homebrew_casks:
       template: "https://github.com/{{ .Env.GH_ORG_NAME }}/container-use/releases/download/{{ .Tag }}/{{ .ArtifactName }}"
     homepage: "https://github.com/{{ .Env.GH_ORG_NAME }}/container-use"
     description: "Containerized environments for coding agents"
+    completions:
+      bash: "completions/cu.bash"
+      zsh: "completions/cu.zsh"
+      fish: "completions/cu.fish"
     hooks:
       post:
         install: |
+          # remove quarantine xattr (note we don't do anything with signatures yet)
           if system_command("/usr/bin/xattr", args: ["-h"]).exit_status == 0
             system_command "/usr/bin/xattr", args: ["-dr", "com.apple.quarantine", "#{staged_path}/cu"]
           end
--- a/install.sh
+++ b/install.sh
@@ -127,6 +127,18 @@ get_latest_version() {
     curl -s "https://api.github.com/repos/$REPO/releases/latest" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/'
 }
 
+# Show shell completion setup instructions
+show_completion_instructions() {
+    local binary="$1"
+
+    log_info "To enable shell completions, run:"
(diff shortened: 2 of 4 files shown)


install.sh
```diff
@@ -127,6 +127,18 @@ get_latest_version() {
...
 }
 
+# Show shell completion setup instructions
+show_completion_instructions() {
+    local binary="$1"
+
+    log_info "To enable shell completions, run:"
+    echo "  Bash (with bash-completion): $binary completion bash > ~/.local/share/bash-completion/completions/cu"
+    echo "  Bash (unconfigured): echo 'source <($binary completion bash)' >> ~/.bashrc"
+    echo "  Zsh (with compinit and a writable fpath[1]):  $binary completion zsh > \"\${fpath[1]}/_cu\""
```
tiborvass, line 137: Docker adds in zshrc, an fpath as the first entry. But i don't want to add cu's completion to my docker's completion folder.

cwlbraa: 🤔 under some package managers fpath[1] isn't gonna be user-writable either, so unideal in several situations.

can i get a citation for the docker approach?

.goreleaser.yaml
```diff
@@ -56,10 +56,26 @@ homebrew_casks:
     hooks:
       post:
         install: |
+          # remove quarantine xattr (note we don't do anything with signatures yet)
           if system_command("/usr/bin/xattr", args: ["-h"]).exit_status == 0
             system_command "/usr/bin/xattr", args: ["-dr", "com.apple.quarantine", "#{staged_path}/cu"]
           end
 
+          # Install shell completions
+          bash_completion_dir = "#{prefix}/etc/bash_completion.d"
```
aluzzardi: Looks like `goreleaser` as built-in completion thingies?

```yaml
    completions:
      bash: "completions/goreleaser.bash"
      zsh: "completions/goreleaser.zsh"
      fish: "completions/goreleaser.fish"
```
https://github.com/goreleaser/goreleaser/blob/main/.goreleaser.yaml#L234-L237

Generated by a hook here https://github.com/goreleaser/goreleaser/blob/main/.goreleaser.yaml#L11

cwlbraa: oh nice, i saw these but missed the hook technique which uses `go run` to generate them

https://github.com/goreleaser/goreleaser/blob/main/scripts/completions.sh

cwlbraa:
tested homebrew on my fork https://github.com/cwlbraa/homebrew-tap/commits/main/, working correctly and installing cu completions at least for zsh.

if you've already got completions in that brew directory, it throws this:

```
Error: It seems there is already a Zsh Completion at '/opt/homebrew/share/zsh/site-functions/_cu'.
```

but after deleting that file and reinstalling the error goes away.

PR #134, opened 2025-06-26
fix: ci -- run all subpackages

grouville:
When moving all the tests to environment/integration, I forgot to update the CI to actually run all subpackages. It also cleans up some over-engineering on my end at the time.

We now have a single test command, which, by default, runs all the tests


grouville:
found a bug

cwlbraa:
@grouville not necessarily for this PR, but i think if you add -short support [here](https://github.com/dagger/dagger/blame/main/modules/go/main.go#L317) we can just use the module (and maybe get more cache hits)

PR #137, opened 2025-06-26
`cu configure`

kpenfound:
Paired on this with Warp.

An interactive experience to configure your agent with container-use

Currently supports:
- Claude Code
- Goose
- Cursor
- Codex
- Amazon Q

Leaving other agents out for now to get this POC in. Some agents require the user to configure things in a UI instead of an easily editable config file.

The interactive UI uses bubbletea for a nice experience.

This should **add** container use to your existing agent configuration, or create a configuration that enables container use. If container use is already configured, we should not break that. This should be able to run multiple times.

**Bonus**: I ran into some errors with environment_file_read and fixed naming consistency and range validation on this branch


aluzzardi:
There's a couple of features from #105 you may want to consider porting here. Maybe.

- Play nice with existing rule files. Like, append to the rules file and use a markdown marker as delimiter to be re-entrant (e.g. on next run, update only the container-use instructions)
- Agent Auto-Detection with Manual Override: oftentimes we can guess the user's agent (`.cursor`, `.claude`, `CLAUDE.md`, ...) and not ask the user
- Auto-Approve MCP tools

As a nit, there's no need to download any files from `cu configure`, the rules are available in a go embed (`rules.AGENT` or something?)

kpenfound:
@aluzzardi this should play nice with existing rules files now

- auto detection makes sense, should be easy enough to check for local configs. Maybe an additional "auto" option in the interactive flow to update all local configs for container-use?
- I added the auto-approve configs for codex to try it out. I need to double check the configurations for the other agents
- yeah i can definitely embed the rules files, thats a good call

cmd/cu/configure.go
```diff
@@ -0,0 +1,392 @@
...
+	cmd := exec.Command("claude", "mcp", "add", "container-use", "--", cuPath, "stdio")
+	if err := cmd.Run(); err != nil {
+		fmt.Printf("Warning: Could not automatically add MCP server: %v\n", err)
+		fmt.Printf("Please run manually: claude mcp add container-use -- %s stdio\n", cuPath)
+	} else {
+		fmt.Println("✓ Added container-use MCP server to Claude")
+	}
+
+	// Download and append agent rules
+	if err := downloadAgentRules("CLAUDE.md"); err != nil {
```
aluzzardi: No need to download -- these are embedded at `rules.AgentRules`

cmd/cu/configure.go
```diff
@@ -0,0 +1,392 @@
...
+		fmt.Printf("Warning: Could not download agent rules: %v\n", err)
+		fmt.Println("Please run manually: curl https://raw.githubusercontent.com/dagger/container-use/main/rules/agent.md >> CLAUDE.md")
+	} else {
+		fmt.Println("✓ Added agent rules to CLAUDE.md")
+	}
+
+	fmt.Println("\nClaude Code configuration complete!")
+	fmt.Println("To use with trusted tools only:")
+	fmt.Println("claude --allowedTools mcp__container-use__environment_checkpoint,mcp__container-use__environment_file_delete,mcp__container-use__environment_file_list,mcp__container-use__environment_file_read,mcp__container-use__environment_file_write,mcp__container-use__environment_open,mcp__container-use__environment_run_cmd,mcp__container-use__environment_update")
```
aluzzardi: Would it be possible to backport a system similar to this?

https://github.com/dagger/container-use/pull/105/files#diff-541ebb6fe3a9c40eb580b5e53397aac809183ecc7157e45a1ee002703d78ae1aR31-R45

Basically:

- Grab the tools dynamically from the codebase
- Ideally, write `.claude/settings.local.json` ourselves (after prompting the user) so it's sticky and auto configured

cmd/cu/configure.go
```diff
@@ -0,0 +1,392 @@
...
+	return nil
+}
+
+func configureGoose() error {
+	fmt.Println("Configuring Goose...")
+
+	configPath := filepath.Join(os.Getenv("HOME"), ".config", "goose", "config.yaml")
```
aluzzardi: nit: use `homedir.Expand()`, already used in the rest of the codebase and more robust

cmd/cu/configure.go
```diff
@@ -0,0 +1,392 @@
...
+	if data, err := os.ReadFile(configPath); err == nil {
+		if err := yaml.Unmarshal(data, &config); err != nil {
+			return fmt.Errorf("failed to parse existing config: %w", err)
+		}
+	} else {
+		config = make(map[string]interface{})
+	}
+
+	// Get extensions map
+	var extensions map[string]interface{}
```
aluzzardi: I'm not too sure if it worked well, but I had made a helper to merge configs:

https://github.com/dagger/container-use/pull/105/files#diff-da11c2eb4c3ee3263855341fece229b30582a33c4224546e61f0f37246f39485R185

Maybe not specifically *that* one, but it might be useful to have some kind of helper with that shared logic (especially since we probably need to fine tune error handling, de-dup logic, etc)

cmd/cu/configure.go
```diff
@@ -0,0 +1,392 @@
...
+	if ext, ok := config["extensions"]; ok {
+		extensions = ext.(map[string]interface{})
+	} else {
+		extensions = make(map[string]interface{})
+		config["extensions"] = extensions
+	}
+
+	// Check if container-use already exists
+	if _, exists := extensions["container-use"]; exists {
```
aluzzardi: IMHO this should be re-entrant and keep reconfiguring ...

For instance, when changing the binary from `cu stdio` to `container-use stdio`, re-running this command should fix that

cmd/cu/configure.go
```diff
@@ -0,0 +1,392 @@
...
+		fmt.Println("✓ container-use already configured in Goose")
+		return nil
+	}
+
+	// Add container-use extension
+	extensions["container-use"] = map[string]interface{}{
+		"name":    "container-use",
+		"type":    "stdio",
+		"enabled": true,
+		"cmd":     "cu",
```
aluzzardi: `claude` is using the absolute path whereas this is using just the binary name -- they all should do one or the other

(maybe all `configureXXX()` should take the cmd as argument so we handle this from the caller?)

cmd/cu/configure.go
aluzzardi: Also -- this should be re-entrant so if we update our rules, re-running this will update those rules

I was injecting a "hidden" marker (`<!-- container-use-rules -->`) so configure could find again its old rules and update them: https://github.com/dagger/container-use/pull/105/files#diff-da11c2eb4c3ee3263855341fece229b30582a33c4224546e61f0f37246f39485R103-R140

cmd/cu/configure.go
```diff
@@ -0,0 +1,392 @@
...
+	Args    []string          `toml:"args"`
+	Env     map[string]string `toml:"env"`
+}
+
+var configureCmd = &cobra.Command{
+	Use:   "configure [agent]",
+	Short: "Configure MCP server for different agents",
+	Long:  `Setup the container-use MCP server according to the specified agent including Claude Code, Goose, Cursor, and others.`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) == 0 {
```
aluzzardi: we should probably open the repository at this point (to make sure we're actually in a repo) and also get the `repo.SourcePath()` and make sure all commands are relative to this path?

e.g. it's `filepath.Join(repo.SourcePath(), ".cursor", ...)` rather than just `.cursor`

aluzzardi (commented):
Nice! Left a bunch of minor comments

Also, question: should this be `cu configure` or `cu mcp configure` or `cu agent configure`? Concerned about "giving away" the top-level `configure` command which could be used to configure the environment itself?

cmd/cu/configure.go
kpenfound: in that case if the user has modified the container-use mcp config we'd reset it to default?

cmd/cu/configure.go
kpenfound: We also may want to let the user select between local and global. `./.cursor` vs `~/.cursor`

aluzzardi:
Naming:

`cu config agent` makes sense to me (no strong opinion though)

not sure about `cu config env` since it (at least in the current implementation) has a bunch of sub-commands: e.g. `base-image set`, `env set` ... and `cu config env env set` would be weird :D 

But:

- `cu config agent` <-- this PR
- `cu config base-image set` <-- the other PR
- `cu config secret add` <-- soon to be PR
- `cu config` <-- interactive

Makes sense I think?

kpenfound:
@aluzzardi agreed, I'll move the `configure` command and add an interactive entrypoint for `cu config`

cwlbraa:
can we get some sort of manual test harness for this? like maybe something that plops out a gitignored directory full of freshly configured empty git repos and best-effort opens a bajillion agent apps?

(trying to be productive about the fact i hate how untestable this is lol)

kpenfound:
Added tests and moved `cu configure [agent]` to `cu config agent [agent]`

aluzzardi (approved):

PR #143, opened 2025-06-27
environment_run: improve prompts for better accuracy

aluzzardi:


mcpserver/tools.go
```diff
@@ -412,7 +412,7 @@ var EnvironmentListTool = &Tool{
 
 var EnvironmentRunCmdTool = &Tool{
 	Definition: mcp.NewTool("environment_run_cmd",
-		mcp.WithDescription("Run a terminal command inside the environment."),
+		mcp.WithDescription("Run a terminal command inside a NEW container within the environment."),
```
cwlbraa, line 415: lol NEW (are the caps really necessary?) (obviously a nit, just curious)

aluzzardi: I asked claude to tweak its own tool prompts so it won't fail next time, that's what it had to say :P

aluzzardi: side quest: I see now it's less confused about life in general, for instance it doesn't attempt to do `cd foo/` and work from there in follow up commands (and rather does `cd foo/ && cmd` for all runs)

PR #149, opened 2025-06-30
dagger: use existing dagger engine if any

### LINKED ISSUE #148: Engine version conflicts between `cu` and `dagger`.
# Problem

When running `cu` as an MCP server, since it's using the Dagger SDK, it'll automatically provision an engine for that SDK. When calling `cu terminal` outside an existing dagger session, we're  calling (https://github.com/dagger/container-use/blob/main/cmd/cu/terminal.go#L39) the `dagger` CLI which if it uses a different version than the one provisioned by cu, will trigger and engine replace by default. 

 In summary, `cu` and `dagger` binaries mismatching is hurting users given that they find themselves having unwanted behaviors and strange issues. 

# Solution

Make `cu` and `dagger` place nice together by allowing `cu` to be able to use whatever engine you have provisioned  as long as it's compatible.

tiborvass:
Fixes #148

### DIFF
--- a/cmd/cu/terminal.go
+++ b/cmd/cu/terminal.go
@@ -36,7 +36,7 @@ var terminalCmd = &cobra.Command{
 				}
 				return fmt.Errorf("failed to look up dagger binary: %w", err)
 			}
-			return syscall.Exec(daggerBin, append([]string{"dagger", "run"}, os.Args...), os.Environ())
+			return syscall.Exec(daggerBin, append([]string{"dagger", "run"}, os.Args...), append(os.Environ(), "DAGGER_LEAVE_OLD_ENGINE=1"))
 		}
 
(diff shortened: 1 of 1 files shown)


kpenfound (approved):
ship it! to confirm my answer to @marcosnils 's question: its ok if the cu mcp is running on a different engine than `cu terminal`?

marcosnils:
> Fixes #148

as discussed in Discord, it doesn't address #148 directly as there's still going to be a struggle between the `cu` and the `dagger` binary re-provisioning engines every time they're called independently of the `cu terminal` sub-command.

aluzzardi (approved):

marcosnils:
@aluzzardi  this has been moved to draft per my last comment :pray:

PR #157, opened 2025-06-30
test: dedup test framework

grouville:
The test framework was based on some copy-pasted code between the MCP tool handlers and the test helpers. This created maintenance burden as any change to the MCP tools required updating the test helpers to match.

This PR introduces two key improvements:

  1. **MCPToolInvoker**: A test helper that directly calls MCP tool handlers instead of duplicating their logic. This ensures tests always use the exact same code paths as real MCP clients, reducing drift and maintenance overhead.

  2. **Context-based test isolation**: Tests now pass `container_use_base_path` via context to isolate each test's container-use data. This allows tests to run in parallel without interference and removes the need for the `test_repository` workaround.

These changes make the test framework more maintainable and reliable while keeping the same clean test API.

### Update 07/08/2025

- Instead of introducing `MCPToolInvoker`, to call the MCP server, I extracted the core logic of the `mcp/tools` handler so that they can be shared with the tests
- A lint test to avoid parallel tests has been introduced, as it is not concurrent safe atm

### DIFF
--- a/mcpserver/tools.go
+++ b/mcpserver/tools.go
@@ -219,7 +219,7 @@ var EnvironmentOpenTool = &Tool{
 			return nil, err
 		}
 
-		dag, ok := ctx.Value("dagger_client").(*dagger.Client)
+		dag, ok := ctx.Value(daggerClientKey{}).(*dagger.Client)
 		if !ok {
 			return mcp.NewToolResultErrorFromErr("dagger client not found in context", nil), nil
 		}
@@ -409,7 +409,7 @@ Supported schemas are:
 			return nil, err
 		}
 
-		dag, ok := ctx.Value("dagger_client").(*dagger.Client)
+		dag, ok := ctx.Value(daggerClientKey{}).(*dagger.Client)
 		if !ok {
 			return mcp.NewToolResultErrorFromErr("dagger client not found in context", nil), nil
 		}
@@ -547,7 +547,7 @@ Failure to do so will result in the tool being stuck, awaiting for the command t
 		useEntrypoint := request.GetBool("use_entrypoint", false)
 		explanation := request.GetString("explanation", "")
 
-		dag, ok := ctx.Value("dagger_client").(*dagger.Client)
+		dag, ok := ctx.Value(daggerClientKey{}).(*dagger.Client)
 		if !ok {
 			return mcp.NewToolResultErrorFromErr("dagger client not found in context", nil), nil
 		}
@@ -674,7 +674,7 @@ var EnvironmentFileReadTool = &Tool{
 			return nil, err
 		}
 
-		dag, ok := ctx.Value("dagger_client").(*dagger.Client)
+		dag, ok := ctx.Value(daggerClientKey{}).(*dagger.Client)
 		if !ok {
 			return mcp.NewToolResultErrorFromErr("dagger client not found in context", nil), nil
 		}
@@ -745,7 +745,7 @@ var EnvironmentFileListTool = &Tool{
 			return nil, err
 		}
 
-		dag, ok := ctx.Value("dagger_client").(*dagger.Client)
+		dag, ok := ctx.Value(daggerClientKey{}).(*dagger.Client)
 		if !ok {
 			return mcp.NewToolResultErrorFromErr("dagger client not found in context", nil), nil
 		}
@@ -820,7 +820,7 @@ var EnvironmentFileWriteTool = &Tool{
 			return nil, err
 		}
 
-		dag, ok := ctx.Value("dagger_client").(*dagger.Client)
+		dag, ok := ctx.Value(daggerClientKey{}).(*dagger.Client)
 		if !ok {
 			return mcp.NewToolResultErrorFromErr("dagger client not found in context", nil), nil
 		}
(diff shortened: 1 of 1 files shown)


repository/repository.go
```diff
@@ -43,7 +43,12 @@ func (r *Repository) getWorktreePath() string {
 }
 
 func Open(ctx context.Context, repo string) (*Repository, error) {
-	return OpenWithBasePath(ctx, repo, cuGlobalConfigPath)
+	basePath := cuGlobalConfigPath
+	// Check if a test base path is provided in context
+	if testBasePath, ok := ctx.Value("container_use_base_path").(string); ok {
```
tiborvass: I'd be okay merging this for now, but in a follow up I think i would rather not have this in the main code path of the `cu` binary if this is only for tests. Maybe make:

```go
var getGlobalConfigPath = func(context.Context) string { return cuGlobalConfigPath }
```

That would be overwritten by the test framework with this ctx logic.

environment/integration/helpers.go
```diff
@@ -48,6 +53,9 @@ func WithRepository(t *testing.T, name string, setup RepositorySetup, fn func(t
 	configDir, err := os.MkdirTemp("", "cu-test-config-"+name+"-*")
 	require.NoError(t, err, "Failed to create config dir")
 
+	// Set the base path in context for all operations in this test
+	ctx = context.WithValue(ctx, "container_use_base_path", configDir)
```
cwlbraa: seems weird to pass this on context given that filesystem operations don't take ctx as an argument.... like if you wanted to unit test those fs operations and they had a dependency on this path, you'd need to inject this through to where the base path gets pulled out

cwlbraa: i read more: tibor's suggestion is good

environment/integration/helpers.go
```diff
@@ -182,93 +190,296 @@ type UserActions struct {
...
-	err = u.repo.Update(u.ctx, env, "Update env "+env.ID, explanation)
-	require.NoError(u.t, err, "repo.Update after UpdateConfig should succeed")
+	
+	result, err := u.mcp.CallTool("environment_update", params)
+	require.NoError(u.t, err, "UpdateEnvironment should succeed")
+	require.NotNil(u.t, result, "UpdateEnvironment should return a result")
 }
 
 // FileDelete mirrors environment_file_delete MCP tool behavior
 func (u *UserActions) FileDelete(envID, targetFile, explanation string) {
```
cwlbraa, line 245: don't necessarily go implementing this, but how much would it suck to just have a generic wrapper around u.mcp.CallTool? needing to maintain these wrappers as tools change seems like a bit of an undue burden to me.

cwlbraa: oh interesting that's already a generic wrapper written by us... what i'm trying to ask is whether it could it be taken far enough to make it usable under test without these extra methods?

tiborvass: Now that i think of it, the logic in mcpserver/tools.go should be MCP boilerplates to Go functions. Those functions could be used directly by the tests.

environment/integration/helpers.go
grouville: I implemented your recommended solution. Unfortunately, I was unable to make it concurrent safe for parallel tests. As you both agree on the implementation, I followed it + added a safeguard on the linter.

Given the size of the test suite atm, it's ok not to be parallel

environment/integration/helpers.go
grouville: Done

repository/repository.go
grouville: Have been unable to make this global concurrent safe. As you both agree on the implementation, and in order to advance on this refacto, I followed the recommendation + a linter to protect against parallel tests

repository/repository.go
```diff
@@ -26,6 +26,29 @@ const (
...
+// Note: This pattern doesn't support parallel tests. Tests using SetTestConfigPath
+// must not call t.Parallel().
+//
+// Usage in tests:
+//
+//	cleanup := repository.SetTestConfigPath("/tmp/test-config")
+//	defer cleanup()
+func SetTestConfigPath(path string) func() {
+	original := globalConfigPath
+	globalConfigPath = func(context.Context) string { return path }
```
aluzzardi, line 48: is the goal to eventually store this in `context` for parallelism?

grouville: That was my first implementation, which was changed following the comments. But yes ✅

aluzzardi: Got it! Yeah sounds good! This looks great, now they're proper integration tests!

.dagger/main.go
```diff
@@ -87,10 +87,58 @@ func (m *ContainerUse) Test(ctx context.Context,
...
+done
+
+if [ "$found_issues" = true ]; then
+  echo ""
+  echo "Tests using WithRepository must not call t.Parallel() as it causes race conditions."
+  exit 1
+fi
+
+echo "✓ No parallel test issues found"
+`
```
cwlbraa, line 134: seems like overkill to me, ngl, but fine to leave in

environment/integration/helpers.go
cwlbraa: yep, fine to figure it out later. tbh i'd rather have parallel tests than not have this issue with context injection in unit tests, i was not thinking about thread safety when i recommended the global fn

cwlbraa (approved):

.dagger/main.go
grouville: Happy to remove it after we make it thread safe 🙏 -- it's just a safeguard against the current open PRs that would definitely be flaky (happened during a rebase)

Issue #110, opened 2025-06-19
cu delete <container> does not work


malkomalko:
On a mac. Deleting a container actually adds a new random container. Also, if I manually delete everything in `~/.config/container-use` it still shows.

```
>  cu version
cu version 0.0.5
commit: e3aabe8daffbe3e12a1042e75848d41657183d98
built: 2025-06-16T23:26:03Z
```

```
>  docker ps
CONTAINER ID   IMAGE                                    COMMAND                  CREATED        STATUS                    PORTS                                           NAMES
258f318dc1cd   registry.dagger.io/engine:v0.18.10       "dagger-entrypoint.s…"   27 hours ago   Up 15 minutes                                                             dagger-engine-v0.18.10
```

```
>  cu list
rm-feat/deals
rm-feat/modern-corgi
rm-feat/popular-sturgeon
```

```
>  cu delete rm-feat/modern-corgi
Creating new Engine session... OK!
Establishing connection to Engine... 1   : /Users/someuser/Library/Caches/dagger/dagger-0.18.10 session --label dagger.io/sdk.name:go --label dagger.io/sdk.version:0.18.10
2   : ┆ connect
3   : ┆ ┆ starting engine
4   : ┆ ┆ ┆ create
5   : ┆ ┆ ┆ ┆ exec docker ps -a --no-trunc --filter name=^dagger-engine-|^dagger-engine-v0\.18\.10$ --format {{.Names}}
5   : ┆ ┆ ┆ ┆ [0.0s] | dagger-engine-v0.18.10
5   : ┆ ┆ ┆ ┆ exec docker ps -a --no-trunc --filter name=^dagger-engine-|^dagger-engine-v0\.18\.10$ --format {{.Names}} DONE [0.0s]
6   : ┆ ┆ ┆ ┆ exec docker start dagger-engine-v0.18.10
6   : ┆ ┆ ┆ ┆ [0.0s] | dagger-engine-v0.18.10
6   : ┆ ┆ ┆ ┆ exec docker start dagger-engine-v0.18.10 DONE [0.0s]
4   : ┆ ┆ ┆ create DONE [0.1s]
3   : ┆ ┆ starting engine DONE [0.1s]
7   : ┆ ┆ connecting to engine
7   : ┆ ┆ [0.0s] | engine name=258f318dc1cd version=v0.18.10 client=oott8xdjvveu8rgc9gv99vb02
[...]

PR #158, opened 2025-07-01
test: extract CLI core logic into shared operations

grouville:
⚠️ Must be a follow-up of #157 

The CLI commands (delete, list, checkout) had their core logic directly embedded in the cobra command handlers. This made it impossible to test the actual CLI behavior from integration tests without duplicating the implementation.

This PR extracts the core operations into a new `cmd/cli/operations.go` package that can be used by both:
  - The CLI commands themselves
  - The tests relying on this interface

The extraction is straightforward - each operation keeps the same logic but moves to a shared function. This allows tests to call the exact same code that users execute via `cu delete`, `cu list`, etc.

This change is particularly important for testing bugs like #110 where we need to verify the actual CLI behavior, not a test-specific reimplementation.

### DIFF
--- a/cmd/cli/operations.go
+++ b/cmd/cli/operations.go
@@ -1,0 +1,81 @@
+// Package cli contains the core logic for container-use CLI commands.
+package cli
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/dagger/container-use/environment"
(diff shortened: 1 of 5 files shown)


cwlbraa:
there is an alternative here: test the binary instead of only packages.

PR #162, opened 2025-07-01
cmd pkg housekeeping

cwlbraa:
this PR moves stdio to its own file and makes the way we call AddCommand consistent across the cmd package: one command per file in its own init.


aluzzardi:
`stdio.go` is missing

cwlbraa:
_old man shakes fist at git_

cwlbraa:
@aluzzardi fixed

tiborvass (approved):

PR #165, opened 2025-07-01
add windsurf config instructions

kpenfound:


rules/windsurf.mdc
```diff
@@ -0,0 +1,9 @@
+---
+trigger: always_on
```
jpadams, line 2: So Windsurf needs this?

Ideally I'd think we'd use one rules file (that we can update in one place in gist or repo) and drop it at the location/filename needed for each tool, but perhaps that dream is not possible.

jpadams (approved):
👍

rules/windsurf.mdc
kpenfound: Yeah windsurf needs it. Cursor's is similar but different 🙃

PR #166, opened 2025-07-01
docs: Remove `claude` installation step to be consistent with others

aluzzardi:


jpadams (approved):
Consistency is good. Only issue is that depending on how you install, you might call `claude` or might call `npx....`, but agree that can be left as an exercise for the user and they can adjust the invocation as well.

PR #168, opened 2025-07-02
add flag push remote in checkout command 

ernesto27:
This request add a flag  -p in command checkout to allow to push branch created after checkout to your remote origin repository,
this is something that me and my teammates are doing  basically for two reasons.

- Some people  prefer to use a browser to review code (github, bitbucket),  and not use command "cu diff " in terminal.
- sometimes we want to push the branch to allow other people to review code

### DIFF
--- a/cmd/cu/checkout.go
+++ b/cmd/cu/checkout.go
@@ -8,8 +8,8 @@ import (
 )
 
 var checkoutCmd = &cobra.Command{
-	Use:   "checkout <env>",
-	Short: "Switch to an environment's branch locally",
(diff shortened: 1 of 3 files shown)


cwlbraa:
`checkout -p` in git is for `--patch`... aka interactively taking chunks of some files and not others, and i don't think we should shadow that. you also don't need to checkout in order to push, like i'd do `git push origin container-use/fine-griffon:feature-name`.

ernesto27:
oh, i didn't know  that you can do that.
i will close this PR

PR #170, opened 2025-07-02
dagger-in-container-use

cwlbraa:
this PR makes dagger usable inside of a container-use environment via ExperimentalPrivilegedNesting.

to make this work, I had to clean up environment export, moving it to repository so we can make sure the git-worktree-pointer `.git` file does _not_ exist inside the container, or else dagger chokes. 

as a follow-up, we're very ready at this point to excise all knowledge of worktrees from the environment package.

### DIFF
--- a/test.txt
+++ b/test.txt
@@ -1,1 +1,0 @@
-test file


test.txt
```diff
@@ -0,0 +1 @@
+test file
```
aluzzardi: forgotten?

.container-use/AGENT.md
```diff
@@ -1,26 +1,27 @@
 This is a development environment for container-use, a CLI tool that provides containerized environments for coding agents.
```
aluzzardi, line 1: mega nit / question: we're kinda hijacking this file for generic agent rules. The pattern we'd advocate for users would be to put all this stuff into `CLAUDE.md` / `.goosehints`, PLUS our environment-specific prompt, right?

And ideally this file will go away soon since it clashes a bit with the existing ones, I guess

environment/environment.go
```diff
@@ -242,8 +225,9 @@ func (env *Environment) Run(ctx context.Context, command, shell string, useEntry
 		args = []string{shell, "-c", command}
 	}
 	newState := env.container().WithExec(args, dagger.ContainerWithExecOpts{
-		UseEntrypoint: useEntrypoint,
-		Expect:        dagger.ReturnTypeAny, // Don't treat non-zero exit as error
+		UseEntrypoint:                 useEntrypoint,
+		Expect:                        dagger.ReturnTypeAny, // Don't treat non-zero exit as error
```
aluzzardi, line 229: this will swallow errors, no? environment_update([`apk add WRONG`]) will happily succeed, no information returned to the LLM

repository/git.go
```diff
@@ -177,6 +178,27 @@ func (r *Repository) propagateToWorktree(ctx context.Context, env *environment.E
 	return nil
 }
 
+func (r *Repository) exportEnvironment(ctx context.Context, env *environment.Environment) error {
+	worktreePointer := fmt.Sprintf("gitdir: %s/worktrees/%s", r.forkRepoPath, env.ID)
+
+	_, err := env.Workdir().
+		WithNewFile(".git", worktreePointer).
```
aluzzardi, line 185: micro nit: I think this is the first time we're crossing the boundary of exposing dagger outside of `environment`.

Couldn't we achieve the same with a dumb `os.WriteFile` here and leaving `Export()` as it is?

.container-use/AGENT.md
cwlbraa: I'm still on the fence about removing `.container-use/AGENT.md`. Right now I'm not using any editor-specific prompt injection files and I'm getting pretty good results by hiding built-in tools, and i think that's thanks to this file...

environment/environment.go
cwlbraa: Good catch, yeah this was broken in a previous PR. Will fix.

test.txt
cwlbraa: Yep llm slop I think, I have 0 recollection of this file.

repository/git.go
cwlbraa: I don't think it is, otherwise there would be new imports? Will check in a bit.

.container-use/AGENT.md
cwlbraa: part of the point of this PR is that we should dogfood this before we remove it

environment/environment.go
cwlbraa: oh wait, no, this is not part of environment_update, this is part of run_command, and the errors are in fact surfaced below.

repository/git.go
cwlbraa: yeah, we're also using dagger in Repository.Create, that boundary was broken when i did "real container resumes". 

and frankly i think breaking that boundary is kinda necessary because we use dagger to manipulate host disk and I don't want environment to know about host disk.

even if we removed the direct client calls from Repository, we're still using the client type as an explicit parameter to pass down to Environment, specifically Repository.Create and Repository.Get require a dagger client.

environment/environment.go
aluzzardi: I misread the diff!

aluzzardi (approved):

PR #171, opened 2025-07-02
tidy up environment logs

aluzzardi:
- Notes are added *after* the commit, instead of being applied to the
  previous commit.
- Cleaner commit messages containing only the explaination. The
  "operation" part (e.g. write file foo) is a note.
- Cleaned up command execution notes
- More concise `cu log` output


environment/integration/integration_test.go
```diff
@@ -93,10 +92,8 @@ func TestEnvironmentIsolation(t *testing.T) {
 		devLog, _ := repository.RunGitCommand(context.Background(), dev.Worktree, "log", "--oneline", "-2")
 		stagingLog, _ := repository.RunGitCommand(context.Background(), staging.Worktree, "log", "--oneline", "-2")
 
-		assert.Contains(t, devLog, "Write config.json")
-		assert.Contains(t, devLog, "Create env")
-		assert.Contains(t, stagingLog, "Write config.json")
-		assert.Contains(t, stagingLog, "Create env")
+		assert.Contains(t, devLog, "Dev config")
+		assert.Contains(t, stagingLog, "Staging config")
```
cwlbraa, line 96: Maybe we wanna get a bit stricter in the tests about not having extraneous audit lines?

PR #175, opened 2025-07-02
Don't commit uncommitted changes, try and warn the user instead.

aluzzardi:
This is an experiment to play around the idea of not committing uncommitted work.

### DIFF
--- a/mcpserver/tools.go
+++ b/mcpserver/tools.go
@@ -252,7 +252,28 @@ DO NOT manually install toolchains inside the environment, instead explicitly ca
 			return mcp.NewToolResultErrorFromErr("failed to create environment", err), nil
 		}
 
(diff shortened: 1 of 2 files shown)


cwlbraa (approved):
i like it, let's try it. i think there's more code you can clean up, though.

cwlbraa:
could also use a test (since it's replacing some complicated shit that definitely should've been tested lol)

aluzzardi:
This is what it looks like:

Screenshot 2025-07-03 at 10 26 34 AM

aluzzardi:
> i like it, let's try it. i think there's more code you can clean up, though.

Surprisingly, no. The only thing it was calling was `commitWorktreeChanges`, and that's used elsehwere

PR #176, opened 2025-07-02
Environment Summary

aluzzardi:
This would basically be the equivalent of a commit message, more
descriptive than the title.

We'd use this for `cu merge`

### DIFF
--- a/mcpserver/tools.go
+++ b/mcpserver/tools.go
@@ -121,6 +121,7 @@ func init() {
 type EnvironmentResponse struct {
 	ID              string                 `json:"id"`
 	Title           string                 `json:"title"`
+	Summary         string                 `json:"summary"`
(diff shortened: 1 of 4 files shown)


mcpserver/tools.go
```diff
@@ -280,6 +290,10 @@ var EnvironmentUpdateTool = &Tool{
 			mcp.Description("Short description of the work that is happening in this environment."),
 			mcp.Required(),
 		),
+		mcp.WithString("summary",
+			mcp.Description("Summary of all the changes that happened in this environment. This should be formatted like a commit message and always kept up to date `environment_update`."),
+			mcp.Required(),
+		),
```
cwlbraa, line 296: instructions, title, base_image, setup commands, envs, and secrets need to be optional if this is required, and if we're going that far, maybe everything on here should be optional.

PR #178, opened 2025-07-03
mcp: add log and diff tools

aluzzardi:
- Add environment log and diff tools via MCP
- Add `-b` flag to `cu {log,diff}`


cmd/cu/diff.go
```diff
@@ -8,17 +8,21 @@ import (
...
-	Example: `# See what changes the agent made
+	Example: `# See what changes the agent made (full diff)
 cu diff fancy-mallard
 
+# Compare against main branch
+cu diff fancy-mallard -b main
+
 # Quick assessment before merging
-cu diff backend-api`,
+cu diff backend-api -b main`,
```
cwlbraa, line 25: in raw git, isn't this equivalent to `git diff main fancy-mallard`? or `git diff fancy-mallard main` we should make sure we're copying that UX wherever possible.

also why does it say backend-api, those are adverb animals lol

aluzzardi: `cu diff` is slightly different than a `git diff` ...  out of the box, it does a `git diff $(git merge-base $(git branch --show-current) CU_ENV))..CU_ENV`

The `-b` here is replacing the `merge-base` stuff with whatever is in `-b`

Really unsure whether we should copy the syntax or not here, especially since I think we'll want to support paths (like git) and the syntax gets tricky

```
git diff main fancy-mallard
git diff main..fancy-mallard
git diff main..fancy-mallard ./cmd
git diff main fancy-mallard -- ./cmd
```

aluzzardi: Any opinion from a git UX POV?

awdemos:
Superseded by #363, which rebases the same functionality onto the current main branch.

awdemos:
This PR is superseded by #363, which rebases the same environment_log / environment_diff tools onto the current main branch. A maintainer can close this one.

PR #183, opened 2025-07-03
docs: add zed agent profile instructions

cwlbraa:
isolating the agent to just container-use is pretty powerful :) 

we will need to remember to update the docs when we add new tools, though, which is quite annoying, ntm users won't necessarily re-paste config later (or re-run `cu configure`)...

### DIFF
--- a/docs/agent-integrations.mdx
+++ b/docs/agent-integrations.mdx
@@ -188,11 +188,60 @@ Then configure the Container Use MCP server in the Zed `settings.json`. Provide 
 }
 ```
 
-Next open the Zed Agent Panel ✨ in the lower right and prompt away!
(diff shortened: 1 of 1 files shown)


aluzzardi:
> or re-run cu configure

well ... if we make the plumbing of `cu configure` more accessible, we could warn the user on `repository_open` (just like uncommitted files)

jpadams (approved):
There are some other Zed tools that might be okay like [`web_search`](https://zed.dev/docs/ai/tools?highlight=fetch#web_search)

There are values for 
```
"default_profile": "container-use",
"always_allow_tool_actions": true,
```
that might get set on their own

PR #184, opened 2025-07-03
fix lint errors

aluzzardi:


mcpserver/tools.go
```diff
@@ -239,7 +241,7 @@ DO NOT manually install toolchains inside the environment, instead explicitly ca
 			return nil, err
 		}
 
-		dag, ok := ctx.Value("dagger_client").(*dagger.Client)
+		dag, ok := ctx.Value(daggerClientKey{}).(*dagger.Client)
```
cwlbraa, line 244: i didn't even know you could do this, you learn something new every day

PR #185, opened 2025-07-03
cli: use charmbracelet/fang

aluzzardi:
Not sure what to think of this ...

New project from charmbracelet (makers of bubbletea, e.g. what we use in the dagger TUI). This is supposed to be:

> The CLI starter kit. A small, experimental library for batteries-included [Cobra](https://github.com/spf13/cobra) applications.

It wraps our `rootCmd` and adds some nice things around it:

Screenshot 2025-07-03 at 4 11 04 PM

Screenshot 2025-07-03 at 4 11 29 PM

/cc @kpenfound who found this


.goreleaser.yaml
```diff
@@ -55,6 +57,7 @@ homebrew_casks:
       template: "https://github.com/{{ .Env.GH_ORG_NAME }}/container-use/releases/download/{{ .Tag }}/{{ .ArtifactName }}"
     homepage: "https://github.com/{{ .Env.GH_ORG_NAME }}/container-use"
     description: "Containerized environments for coding agents"
+    manpage: man/container-use.1
```
aluzzardi, line 60: /cc @cwlbraa ... why not?

aluzzardi:
For some reason ... `cu stdio` hangs with this after Ctrl-C

I tried to re-use the same context (https://github.com/dagger/container-use/pull/188) but that doesn't fix this

**edit**: Paging @tiborvass who's an expert at fixing weird hangs :P 

**edit**: Found the issue and patched

.goreleaser.yaml
cwlbraa: if it works, it works... idk what it's generating tbh, but i like the concept. you might wanna cut the next release though to debug ;)

cmd/container-use/main.go
```diff
@@ -24,20 +25,33 @@ Each environment runs in its own container with dedicated git branches.`,
...
 	if err := setupLogger(); err != nil {
 		fmt.Fprintf(os.Stderr, "Failed to setup logger: %v\n", err)
 		os.Exit(1)
 	}
 
-	if err := rootCmd.ExecuteContext(ctx); err != nil {
-		fmt.Fprintf(os.Stderr, "%v\n", err)
+	// FIXME(aluzzardi): `fang` misbehaves with the `stdio` command.
+	// It hangs on Ctrl-C. Traced the hang back to `lipgloss.HasDarkBackground(os.Stdin, os.Stdout)`
+	// I'm assuming it's not playing nice the mcpserver listening on stdio.
```
cwlbraa, line 40: this seems like something @vito might know something about

cmd/container-use/main.go
```diff
@@ -24,20 +25,33 @@ Each environment runs in its own container with dedicated git branches.`,
 )
 
 func main() {
+	ctx := context.Background()
```
cwlbraa, line 28: this is weird, do all the non-stdio commands still get signals?

cwlbraa (approved):

cmd/container-use/main.go
vito: Ah yeah this is a common hangup I've run into. It tries to determine if the terminal has a light or dark background by sending a special sequence on `stdout` and waiting for a reply on `stdin`, which of course not everything is equipped to do. It makes sense to bypass all that for `stdio`.

cmd/container-use/main.go
aluzzardi: It's handled by `fang.WithNotifySignal(os.Interrupt, os.Kill, syscall.SIGTERM)`

PR #193, opened 2025-07-04
allow to pick an environment by default

eunomie:
In case there's one single environment available and none is set as an argument, pick the only environment.

So that it's possible to run `cu merge`, `cu diff`, `cu terminal` without arguments if there's a single environment.

### DIFF
--- a/cmd/cu/env.go
+++ b/cmd/cu/env.go
@@ -1,0 +1,30 @@
+package main
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/dagger/container-use/repository"
+)
+
+func envOrDefault(ctx context.Context, arg string, repo *repository.Repository) (string, error) {
+	if arg != "" {
+		return arg, nil
+	}
+	if list, err := repo.List(ctx); err != nil {
+		return "", err
+	} else if len(list) == 0 {
+		return "", fmt.Errorf("no environment found")
+	} else if len(list) == 1 {
+		return list[0].ID, nil
+	} else {
+		return "", fmt.Errorf("please specify an environment")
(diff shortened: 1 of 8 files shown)


aluzzardi (approved):

cmd/cu/env.go
```diff
@@ -0,0 +1,34 @@
...
+		return arg, nil
+	}
+	if list, err := repo.List(ctx); err != nil {
+		return "", err
+	} else if len(list) == 0 {
+		return "", fmt.Errorf("no environment found")
+	} else if len(list) == 1 {
+		return list[0].ID, nil
+	} else {
+		return "", fmt.Errorf("please specify an environment")
```
aluzzardi, line 25: nit: Should we output a `cu list`?

aluzzardi:
@eunomie Actually, if you don't mind I'd like to put this on hold for a second.

I *think* we *may* want to attach environments to branches and reserve the "argument-less" command for that (e.g. `git checkout foo && cu terminal`).

Not sure just yet, happy to chat about that -- but I think it'd be better to hit pause for a second so we leave the possibility open

eunomie:
> @eunomie Actually, if you don't mind I'd like to put this on hold for a second.

No problem at all. This PR was mostly about the DX, the ability to just `cu something` without to write the environment.
But the branch - environment thing sounds good to me, so let's wait a bit until we decide how we want to solve this.

cwlbraa:
superceded by #224 - a more nuanced version of this that will only give envs descending from the current HEAD as well as adding a selector prompt when that's ambiguous

PR #195, opened 2025-07-07
Docs split integrations

shykes:
Split up integrations docs so that each integration gets its own page, visible in the navbar.

### DIFF
--- a/docs/amazon-q.mdx
+++ b/docs/amazon-q.mdx
@@ -1,0 +1,84 @@
+---
+title: Amazon Q Developer
+description: "Setup guide for Amazon Q Developer with Container Use"
+icon: robot
+---
+
+## Add MCP Configuration
+
+Add this configuration to `~/.aws/amazonq/mcp.json`:
+
+```json
+{
+  "mcpServers": {
+    "container-use": {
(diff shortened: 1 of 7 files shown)


cwlbraa (approved):

cwlbraa:
@shykes why not split out all of the agents? like our docs are small enough for right now that all the agents in "other agents" could go in the sidebar too

aluzzardi:
The current agents page is linked in a few places, we need to figure out how to do linking with this approach (e.g. in the quickstart there's tabs for popular agents and then a link to all agents, there's another link in the landing page, etc).

Potential solutions:

- Leave everything as a single page, BUT add the individual agents we want to the navbar via anchors (e.g. Hopefully `agent-ingrations.mdx#Cursor` works?)
- Create an index file and link to that

shykes:
> @shykes why not split out all of the agents? like our docs are small enough for right now that all the agents in "other agents" could go in the sidebar too

Ha ha I vibe-coded this and didn't notice :-P So yes I agree

jpadams:
I'm also interested in something like this. Will take a closer look.

PR #198, opened 2025-07-07
env: add new FileSystem-related tools such as Grep, Glob and Edit.

tiborvass:
Also modifies List to add an `ignore` argument.


environment/filesystem.go
```diff
@@ -60,3 +78,49 @@ func (env *Environment) FileList(ctx context.Context, path string) (string, erro
 	}
 	return out.String(), nil
 }
+
+func (env *Environment) FileGrep(ctx context.Context, path, pattern, include string) (string, error) {
+	// Hack: use busybox to run `sed` since dagger doesn't have native file editing primitives.
+	args := []string{"/bin/grep", "-E", "--", pattern, include}
```
aluzzardi: What about using `ripgrep` instead? It's smart enough to avoid grepping through `node_modules` and other crap

environment/filesystem.go
```diff
@@ -2,10 +2,18 @@ package environment
...
 	"fmt"
 	"strings"
+
+	"dagger.io/dagger"
+	"dagger.io/dagger/dag"
+	"github.com/dagger/container-use/edit"
 )
 
+// FIXME: See hack where it's used
+const fileUtilsBaseImage = "busybox"
```
aluzzardi: nit: I think there's another one of this pointing to `alpine` in the codebase, can we merge the two?

nit2: can we pin this to a digest? Otherwise AFAIK buildkit will ping the registry each time to resolve the digest and see if it changed compared to local

cmd/cu/stdio.go
```diff
@@ -30,10 +32,15 @@ var stdioCmd = &cobra.Command{
 		}
 		defer dag.Close()
 
+		go warmCache(ctx, dag)
```
aluzzardi: sneaky ;)

aluzzardi (commented):
Looks good to me!

I'll let @vito and @kpenfound do a review -- not too familiar with what other agents tools do (line based edit vs search and replace etc)

environment/filesystem.go
vito: Claude choked on this command when I gave it a go:

https://dagger.cloud/vito/traces/b72d0edd3100fb0789cb2f718eba3b0e?span=af8472a10106d33f

Looks like `grep` didn't like a blank `include` arg. Maybe it should default to `.`?

mcpserver/tools.go
```diff
@@ -612,9 +629,163 @@ var EnvironmentFileListTool = &Tool{
...
+		mcp.WithString("environment_id",
+			mcp.Description("The ID of the environment for this command. Must call `environment_create` first."),
+			mcp.Required(),
+		),
+		mcp.WithString("target_file",
+			mcp.Description("Path of the file to edit, absolute or relative to the workdir."),
+			mcp.Required(),
+		),
+		mcp.WithArray("edits",
+			mcp.Description("Array of sed search-replace operations to perform on the contents of target_file (e.g. \"s/old/new/g\").\nUses extended regex syntax."),
```
vito: Claude choked on this too:

```
● container-use - environment_file_edit (MCP)(environment_source: "/var/home/vito/src/bind", environment_id: "oriented-deer", target_file: "treesitter/binding.gyp", edits:
                                             ["s/tree_sitter_bind_binding/tree_sitter_sprout_binding/g"], explanation: "Updating binding.gyp to use sprout instead of bind")
  ⎿  Error: MCP error -32603: could not bind arguments

● Let me read and update the file content:
```

I think the description is just out of date - looks like it's no longer `sed` syntax (which I think makes sense, since Claude by necessity includes a bunch of context to ensure it doesn't misfire edits and that'd be awkward to squish into `sed` syntax).

aluzzardi:
@tiborvass btw -- with @grouville's new integration tests / MCP fan-out, perhaps we can now actually test those (once we're happy with the signatures)?

tiborvass:
@aluzzardi  yes, that's why it was in Draft, wanted to add tests.

vito:
@tiborvass I'm working on a couple of Dagger PRs that may be useful here:

* https://github.com/dagger/dagger/pull/10742 (merged! should ship today)
* https://github.com/dagger/dagger/pull/10747 (working, but prob won't make the cut for release)
* also have a WIP for `File.replace`, using the same fork of `replace` that you're using here

Would have pinged earlier about it, but I just started on these on a whim the other day and it ended up going smoother than I expected. :pray:

tiborvass:
@vito nice thank you! Feel free to take over the branch or steal from it whatever you need, i'm not attached to it!

PR #199, opened 2025-07-07
wip: file search/replace (needs tests)

vito:
This implements find-and-replace in Go. If there are multiple matches for `search_text` it'll return an error with match IDs, which are deterministic hashes to prevent the model from jumping straight to 'replace the first match.'

(Vibe coded with Claude + `container-use`, tested and seems to work, but n=2)


vito:
this was just submitted to compare notes, prob should have just linked the branch! - see #198

PR #202, opened 2025-07-07
cu config: user-defined environment configuration

aluzzardi:


aluzzardi:
@kpenfound I originally tried `cu env` but it felt *wrong*

- Shouldn't (almost) all `cu` commands be `cu env {merge,diff,...}`? 
- env commands (e.g. `cu env show`) don't feel as clear as `cu config show`

Trying `cu config` for now. For `cu configure` I see two options:

- Moving `cu configure` under `cu config` (e.g. `cu config agent`)
- Renaming one or the other: `cu config` + `cu setup`, `cu env` + `cu configure`

docs/secrets.mdx
```diff
@@ -0,0 +1,147 @@
...
+  <Tab title="🐚 Shell">
+    ```bash
+    # Shell scripts can also access secrets
+    curl -H "Authorization: Bearer $API_KEY" https://api.example.com
+    ```
+  </Tab>
+</Tabs>
+
+<Warning>
+  **Security Note**: While your code can access secrets normally, Container Use automatically strips secret values from logs and command outputs. This means `echo $API_KEY` or similar commands won't expose secrets in the development logs that agents or users can see.
```
cwlbraa, line 146: i'm assuming you tested this? or do you just know this to be true based on the implementation of secrets in dagger?

aluzzardi: The latter

PR #203, opened 2025-07-08
ephemeral config changes

aluzzardi:
Experimental PR that moves the ownership of env config to the user.

1) Environment updates are ephemeral, the mcp tool will instruct the agent to inform the user to run `cu config` commands to make the changes persistent.
2) Separate `environment_config` (base image, setup commands) and `environment_update_metadata` (title) into 2 distinct tools.
3) Removed `instructions` from container-use -- updated our own repo to use CLAUDE.md and .goosehints

### DIFF
--- a/.goosehints
+++ b/.goosehints
@@ -1,0 +1,1 @@
+AGENT.md
--- a/docs/environment-configuration.mdx
+++ b/docs/environment-configuration.mdx
@@ -1,35 +1,49 @@
 ---
 title: Environment Configuration
-description: "Configure your project's default environment with custom base images and setup commands. Set up the foundation that all agent environments will start from."
+description: "Configure your project's default environment and understand how agents adapt their environments during work. Manage ephemeral configuration changes made by agents."
 icon: gear
 ---
 
 ## Overview
 
-Environment configuration lets you define the **default environment** that all agents will start from when working on your project. Instead of using the generic Ubuntu environment, you can specify exactly what base image, dependencies, and setup your project needs.
+Environment configuration in Container Use works in two layers:
+
+1. **Default Configuration**: The foundation environment that all agents start from
+2. **Agent Environment Adaptation**: Changes agents make to their environment during work (ephemeral until imported)
+
+Instead of using the generic environment, you can specify exactly what base image, dependencies, and setup your project needs as defaults. Agents can then adapt their environment as they discover additional requirements during their work.
 
 <Note>
-  Configuration applies to **new environments only**. Existing environments will
-  continue using their original setup.
+  **Agent configuration changes are ephemeral** - they only exist within the agent's environment until you explicitly import them using `container-use config import`.
 </Note>
 
-## When to Configure Environments
+## The Configuration Workflow
 
-Consider customizing your environment when:
+<Steps>
+  <Step title="Set Default Configuration">
+    Configure your project's baseline environment that all agents will start from
+  </Step>
+  <Step title="Agent Starts with Defaults">
+    When an agent creates a new environment, it begins with your default configuration
+  </Step>
+  <Step title="Agent Adapts as Needed">
+    During work, the agent may modify its environment configuration - adding tools, changing base images, or setting variables
+  </Step>
+  <Step title="View Agent Changes">
+    Use `container-use config view <env>` to see what configuration changes the agent made
+  </Step>
+  <Step title="Import Useful Changes">
+    Use `container-use config import <env>` to adopt the agent's configuration improvements as your new defaults
+  </Step>
+</Steps>
 
-- **Your project requires a specific language runtime** (Python, Node.js, Go, etc.)
-- **You need specific system packages** or tools installed
-- **You want consistent development dependencies** across all agent sessions
-- **Your project has a complex setup process** that should be automated
+## Default Environment Configuration
 
-<Card title="Default vs Configured" icon="balance-scale">
-  **Default**: Basic Ubuntu 24.04 with essential tools (git, curl, build tools)
+Configure the baseline environment that all agents will start from when working on your project.
 
-**Configured**: Your specified base image + automated setup commands + environment variables
+By default, environments use a basic Ubuntu image with essential tools (git, curl, ...).
 
-</Card>
-
-## Quick Start
+### Quick Start
 
 Set up a Python project environment:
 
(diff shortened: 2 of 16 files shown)


.goosehints
```diff
@@ -0,0 +1 @@
+AGENT.md
```
cwlbraa, line 1: is this how it's showing symlinks now?

mcpserver/tools.go
```diff
@@ -340,55 +365,60 @@ Supported schemas are:
...
+		if len(suggestions) > 0 {
+			message += "YOU MUST tell the user to run these commands to make the environment configuration changes:\n\n"
+			for _, suggestion := range suggestions {
+				message += fmt.Sprintf("  %s\n", suggestion)
+			}
+			message += "\nExplain to the user that after running these commands, future environments will use the updated configuration."
+		} else {
+			message += "If environment configuration changes are needed, inform the user they can run:\n\n"
+			message += "  cu config base-image set <image>     # Set base container image\n"
+			message += "  cu config setup-command add \"<cmd>\"  # Add setup command\n"
```
cwlbraa: this probably needs to do a diff of new and existing setup commands?

mcpserver/tools.go
```diff
@@ -340,55 +365,60 @@ Supported schemas are:
...
-		config.Env = envs
 
-		secrets, err := request.RequireStringSlice("secrets")
-		if err != nil {
-			return nil, err
-		}
-		config.Secrets = secrets
+		// Create directive message for the agent
+		message := `SUCCESS: Configuration successfully applied. Environment has been restarted, all previous commands have been lost.
+The changes are temporary, you MUST inform the user to make the changes persist
```
cwlbraa:
```suggestion
The changes are temporary, you MUST inform the user to make the changes persist for new environments
```

maybe?

environment/config.go
```diff
@@ -70,10 +66,6 @@ func (config *EnvironmentConfig) Save(baseDir string) error {
 		return err
 	}
 
-	if err := os.WriteFile(path.Join(configPath, instructionsFile), []byte(config.Instructions), 0644); err != nil {
-		return err
-	}
```
cwlbraa, line 135: so we stop saving instructions, but should did we also stop saving environment.json? otherwise they're gonna merge it back even though the prompting is all like "this is temporary! run cu config ..."

.goosehints
aluzzardi: Yup

mcpserver/tools.go
aluzzardi: Yeah the prompt is very broken-ish, not optimized, just toying with the idea

environment/config.go
aluzzardi: Weird ... I think I lost part of my changes, I definitely removed it from `exportEnvironment` at some point.

In any case, this wasn't finished: If we go this route, I think we'd have to store the config in `state`, so ephemeral config updates are persisted for environment resuming. This also means we can view the current config, and maybe have tools to save the updated config locally.

Actually, maybe it's me who reverted back this change -- since we always resume environments (even when doing a follow up MCP call), not saving the updated config on disk means we lose the update and the agent gets very confused.

Double actually -- maybe a better UX would be to hint the user to do a `cu config import <env>` or something (or `cu apply --config <env>` or whatnot)

cwlbraa: don't we implicitly save the updated config on disk by saving the container? 

that does leave a hole if you resume and then try to "import" but that's a bit of a hypothetical edge case? although i guess if you've got a freestanding import subcommand it has to resume before it imports.

.goosehints
cwlbraa: awful.

aluzzardi:
Pushed an update: Config now is always in state, loaded from git only once at creation and then persisted.

Removed the suggestions here for now -- the idea would be to replace them with a "merge config command". Right now it can be inspected with the (hidden) `cu inspect` command.

Example: 

```console
$ cat .container-use/environment.json
{
  "setup_commands": [
    "apt-get update \u0026\u0026 apt-get install -y curl git build-essential",
     ...
  ]
}

$ goose run --debug -t "run ./docs using npx mint dev"

$ cu diff stirring-flamingo
[nothing]

$ cu inspect stirring-flamingo
{
  "state": {
    "config": {
      "setup_commands": [
        "curl -fsSL https://deb.nodesource.com/setup_20.x | bash -",
        "apt-get install -y nodejs"
      ]
    },
  },
}
```

- I need to tweak the tool to do the right thing, right now it replaced my setup_commands rather than adding, but you get the idea
- `cu inspect` is not good enough. In #202 I'm adding `cu config` to manipulate the config, including a `cu config view`. Maybe `cu config view [env]` would show the config for a given environment?
- Maybe `cu merge` and `cu apply` should have a flag to bring in config changes (or maybe `cu config apply <env>`?)

docs/environment-configuration.mdx
```diff
@@ -126,7 +281,7 @@ container-use config setup-command add "pip install pytest black flake8"
 container-use config setup-command add "./scripts/dev-setup.sh"
 
 # Multiple commands with &&
-container-use config setup-command add "apt update && apt install -y postgresql-client"
+container-use config setup-command add "apt update && apt install -y curl jq"
```
cwlbraa: this is a weird to thing to document, is there special logic for &&?

docs/environment-configuration.mdx
```diff
@@ -51,6 +79,133 @@ container-use config show
 
 Now all new agent environments will start with Python 3.11, your dependencies pre-installed, and environment variables configured.
 
+## Agent Environment Adaptation
```
cwlbraa, line 68: Maybe i'm missing it, but I really want a chunk of quickstart documentation that shows an example like "look around this repo and set up a dev env with all the appropriate dependencies for developing & testing this codebase. validate that the workflow associated with each dependency can be completed in the environment"

cwlbraa: this is the sexiest part of env config in container use, imo, and if we're gonna talk about configuration in quickstart at all, we should focus on the exciting part where you can make the agent yak shave for you.

mcpserver/tools.go
```diff
@@ -354,55 +314,112 @@ Supported schemas are:
...
+			return mcp.NewToolResultErrorFromErr("failed to update repository", err), err
 		}
 
 		out, err := marshalEnvironment(env)
 		if err != nil {
 			return mcp.NewToolResultErrorFromErr("failed to marshal environment", err), nil
 		}
-		return mcp.NewToolResultText(fmt.Sprintf("Environment %s updated successfully. Environment has been restarted, all previous commands have been lost.\n%s", env.ID, out)), nil
+
+		// FIXME(aluzzardi): `cu config apply` is not implemented yet.
```
cwlbraa: is this still true?

cwlbraa: we just decided to call it import, instead, no?

cwlbraa (approved):
impl lgtm.

docs could use some work, they're way too verbose in the typical way that claude-generated docs are way too verbose, and I think they walk right past the "exciting" quickstart experience.

aluzzardi:
Thanks! I literally just AI generated the docs while I was waiting for the implementation review, they're rough. I'll rework that

PR #206, opened 2025-07-08
chore: bump dagger version

grouville:
potential fix to https://discord.com/channels/707636530424053791/1375159310757134386/1392278947709845514

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.24.3
 toolchain go1.24.4
 
 require (
-	dagger.io/dagger v0.18.11
+	dagger.io/dagger v0.18.12
 	github.com/dustin/go-humanize v1.0.1
 	github.com/dustinkirkland/golang-petname v0.0.0-20240428194347-eebcea082ee0
 	github.com/mark3labs/mcp-go v0.29.0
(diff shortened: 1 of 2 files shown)


aluzzardi:
Thank you @grouville -- I'd like to understand the root cause though, this is not the first time it happens and we bumped version the last time.

Maybe @sipsma has some thoughts?

Context: using the Go SDK directly, when the CLI version is newer than the version the SDK uses, it fails:

```
  ⎿  Error: failed to list directory: returned error 502: {"data":null,"errors":[{"message":"http do: Post \"http://dagger/query\": command [docker exec -i dagger-engine-v0.18.11 buildctl dial-stdio] has exited with exit
     status 1, make sure the URL is valid, and Docker 18.09 or later is installed on the remote host: stderr=Error response from daemon: No such container: dagger-engine-v0.18.11\n"}]}
```

In this case, I had a `dagger-engine-v0.18.12` running but the SDK was using `dagger-engine-v0.18.11` and not installing the extra version for some reason?

grouville:
> Thank you @grouville -- I'd like to understand the root cause though, this is not the first time it happens and we bumped version the last time.
> 
> Maybe @sipsma has some thoughts?
> 
> Context: using the Go SDK directly, when the CLI version is newer than the version the SDK uses, it fails:
> 
> ```
>   ⎿  Error: failed to list directory: returned error 502: {"data":null,"errors":[{"message":"http do: Post \"http://dagger/query\": command [docker exec -i dagger-engine-v0.18.11 buildctl dial-stdio] has exited with exit
>      status 1, make sure the URL is valid, and Docker 18.09 or later is installed on the remote host: stderr=Error response from daemon: No such container: dagger-engine-v0.18.11\n"}]}
> ```
> 
> In this case, I had a `dagger-engine-v0.18.12` running but the SDK was using `dagger-engine-v0.18.11` and not installing the extra version for some reason?

Making a repro on the side, thank you ! 🙏

grouville:
Superseded by #215

cwlbraa:
closed, unclear if this fixes the linked thing but it does fix `cu terminal` cache misses

grouville:
> closed, unclear if this fixes the linked thing but it does fix `cu terminal` cache misses

It will actually, but above link needs a fix in the engine / SDKs 😇 (on it)

Issue #142, opened 2025-06-27
Error: git command failed (exit code 128): exit status 128


surajitkar:
Environment: MAC

Directory Structure:

├── .DS_Store
├── .git
├── .github
├── .gitignore
├── .vscode
├── GEMINI.md
├── README.md
├── SOCIAL_LOGIN_SETUP.md
├── backend
├── chatbot
├── designs
├── docs
└── frontend

cu list
Error: git command failed (exit code 128): exit status 128
Output: fatal: No such remote 'container-use'

Usage:
  cu list [flags]

Flags:
  -h, --help       help for list
      --no-trunc   Don't truncate output
  -q, --quiet      Display only environment IDs

git command failed (exit code 128): exit status 128

Can you help?

grouville:
Hi @surajitkar,

What version of container-use are you using ? Do you have more repro steps ? I can't seem to repro unfortunately.

Before doing the `cu list` ?

grouville:
Can you also please tell me your: `which git` as well as the output of `git --version` please ? 🙏

surajitkar:
Hi @grouville ,

/usr/local/bin/git
git version 2.20.1

cu version 0.1.0
commit: 410ddd56d059fa517f323f0cf12f08fde413b8f7
built: 2025-06-26T20:08:31Z

surajitkar:
Hi @grouville, do you have any updates on this or a workaround?

grouville:
Hi @surajitkar,

I think we found the root cause -- Apple's `git` CLI does not handle a pattern that we have here, for that version (prior 2.24 if my memory is correct): https://github.com/dagger/container-use/blob/main/repository/repository.go#L243.

As this argument is not handled, it fires an exit code 128. I'm thinking about a solution atm, will fire it today 🙏 

One easy workaround is to rely on a newer git / brew's git

surajitkar:
Hi @grouville,

I have updated git but, same error

cu version 0.1.0
commit: 410ddd56d059fa517f323f0cf12f08fde413b8f7
built: 2025-06-26T20:08:31Z
git version 2.50.0

Input
{
  "environment_source": "/Users/suri/Documents/Projects/trader/booking-platform",
  "explanation": "Creating environment to test the Smart Merge System for AvailabilityManager component without committing changes",
  "title": "Smart Merge System Implementation Test"
}

Output
failed to create environment: failed to commit worktree changes: git command failed (exit code 128): exit status 128
[...]

PR #207, opened 2025-07-09
feat: add enhanced diagnostics to cu version --verbose

grouville:
## Summary

This PR enhances `cu version --verbose` to become a comprehensive diagnostic tool that helps debug user issues by capturing system state and highlighting inconsistencies.

## Problem

When users report issues, we often need them to run multiple commands to understand their system state. Even then, we miss critical inconsistencies that reveal bugs. For example:
- Users see "No such container: dagger-engine-v0.18.11" when v0.18.12 is running
- `cu list` shows no environments while directories exist on disk
- Git remote issues that aren't immediately apparent

## Solution

Transform `cu version --verbose` into a diagnostic snapshot tool that:
1. Captures complete system state in one command
2. Highlights inconsistencies between different views (git, filesystem, docker)
3. Presents data without prescriptive advice, letting the facts speak

## Key Improvements

**Before:** Basic version info with verbose explanations
**After:** Complete state snapshot with inconsistency detection

The tool now captures:
- All Dagger engines running (not just first one)
- Dagger SDK version from go.mod 
- Git state including branches, remotes, and notes
- Filesystem state of environments
- Critical environment variables
- Inconsistencies between these states

## Example Output

```
VERSION: dev (darwin/arm64)
BUILD: abc123

GIT:
  in_repo: true
  branch: main
  remotes: 2
  cu_remote: true (reachable: true)
  cu_branches: 3
  worktrees: 4

DOCKER:
  available: true
  dagger_sdk: v0.18.11
  dagger_engines: 1 running
    - dagger-engine-v0.18.12 (v0.18.12)

ENVIRONMENTS:
  my-env:
    worktree: true, cu_dir: true, env.json: true, agent.md: true
    git_branch: true, git_notes: false

INCONSISTENCIES:
  - Dagger SDK v0.18.11 expects matching engine, but found: v0.18.12
  - 'my-env': has worktree but no git notes (invisible to 'cu list')
```

## How This Helps

1. **Version Mismatch Issues**: Immediately shows when Dagger SDK expects different engine version
2. **Missing Environments**: Reveals why environments exist on disk but not in `cu list` 
3. **State Corruption**: Identifies partial states and missing components
4. **One Command**: Users run one command instead of many

## Testing & Feedback Needed

**If you're experiencing container-use issues, please try this out:**
```bash
# Checkout this branch
git fetch origin pull/XXX/head:test-diagnostics
git checkout test-diagnostics
go build -o cu ./cmd/cu

# Run the diagnostic
./cu version --verbose

# Share the output in your issue
```

**We especially want feedback on:**
- Does this capture the information needed to debug your issue?
- Are there other inconsistencies we should detect?
- Is the output format helpful or should we adjust it?

This addresses issues like #142 where users have environments that exist but don't appear in `cu list`.

### DIFF
--- a/cmd/container-use/diagnostics.go
+++ b/cmd/container-use/diagnostics.go
@@ -1,0 +1,502 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"slices"
+	"strings"
+	"time"
+
+	"github.com/dagger/container-use/repository"
+)
+
+// VersionInfo contains version information.
+type VersionInfo struct {
+	Version   string `json:"version"`
+	Commit    string `json:"commit"`
+	BuildDate string `json:"build_date"`
+	GoVersion string `json:"go_version"`
+	OS        string `json:"os"`
+}
+
+// Snapshot captures the complete diagnostic state.
+type Snapshot struct {
+	Version         VersionInfo                `json:"version"`
+	Git             GitInfo                    `json:"git"`
+	Docker          DockerInfo                 `json:"docker"`
+	Filesystem      FilesystemInfo             `json:"filesystem"`
+	Environments    map[string]EnvironmentInfo `json:"environments"`
+	Inconsistencies []string                   `json:"inconsistencies,omitempty"`
+	RecentErrors    []string                   `json:"recent_errors,omitempty"`
+}
+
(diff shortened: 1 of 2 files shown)


grouville:
Closing this PR in favor of a simpler implementation. I'm following up with a new PR that extracts just the system info feature without the complex diagnostic framework.

PR #211, opened 2025-07-09
Initial MCP Roots support

cwlbraa:
impl is as minimal as possible. tested via `echo {json} | timeout 5 cu stdio`, still need to test with actual vscode.

currently only used to facilitate better error messages on repo.Open, but with dynamic tools we can also use the roots to provide better tool descriptions.

### DIFF
--- a/mcpserver/roots.go
+++ b/mcpserver/roots.go
@@ -2,7 +2,6 @@ package mcpserver
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"log/slog"
 	"strings"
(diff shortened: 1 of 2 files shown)


mcpserver/roots.go
```diff
@@ -0,0 +1,100 @@
...
+	case session.NotificationChannel() <- notification:
+		slog.Info("Requested roots/list from client")
+	case <-ctx.Done():
+		return
+	}
+}
+
+// updateClientRoots parses roots from notification params and updates the global clientRoots
+func updateClientRoots(notification mcp.JSONRPCNotification) int {
+	if notification.Params.AdditionalFields != nil {
```
aluzzardi: Silly lint nit: return early to avoid nesting

mcpserver/roots.go
```diff
@@ -0,0 +1,100 @@
...
+		for _, root := range clientRoots {
+			uri := root.URI
+			uri = strings.TrimPrefix(uri, "file://")
+			if root.Name != "" {
+				baseMsg += fmt.Sprintf("\n  - %s (%s)", uri, root.Name)
+			} else {
+				baseMsg += fmt.Sprintf("\n  - %s", uri)
+			}
+		}
+		return errors.New(baseMsg)
```
aluzzardi: nit: This is throwing away the error wrapping (e.g. errors.Is etc won't work after going through this).

Any chance this could be done with a fmt.Errorf()?

aluzzardi (commented):
Looks good, left minor nits

mcpserver/roots.go
cwlbraa: i changed it to this kinda unthinkingly because i had an actual linter complaining about fmt.Errorf(baseMsg). will fix.

mcpserver/roots.go
cwlbraa: oh nice tq, i also hate this and 100% blame claude for having no taste

cwlbraa:
tested this with vscode and i can't get to the error path at all because agent mode auto-injects the root dirs into the context, so this may be useless for right now.

aluzzardi (approved):

awdemos:
Superseded by #361, which rebases the same functionality onto the current main branch.

awdemos:
This PR is superseded by #361, which rebases the same MCP roots support onto the current main branch. A maintainer can close this one.

PR #213, opened 2025-07-10
ci: maintain a release branch pointing to the latest tag

aluzzardi:


cwlbraa (approved):

cwlbraa:
wait, doesn't this also need to change where the mintlify sources content from? or is that managed outside of code?

aluzzardi:
Yeah outside the code, in the mintlify dashboard itself

PR #217, opened 2025-07-11
add `follow` command

vito:
NOTE: we haven't bikeshedded this at all, it was easy to vibecode so just tossing up a PR. Dogfooded while developing and it works well, but there might be a better approach.

The goal here is to approximate the usual Claude Code flow which allows you to observe changes the agent makes directly in your editor, as it makes them. This makes it easier to notice when the agent is going off on a tangent, and go course-correct.

* add `follow` command, which is `container-use checkout` + fsnotify-driven `git pull`
* tweak logging stack so `follow` can log to the user in a readable fashion
  * if stdout is a TTY, pretty logs to `/dev/stderr`
  * else, do current behavior (log file)
* reduce log level of `git` commands

The latter two are a bit opinionated, not sure if load-bearing or fine as-is.

### Alternative approaches

One idea: tell the agent to put an environment in 'foreground mode' which tells the MCP server to do these same things after each change. That way you don't need to run this command in another terminal, it's all just self-contained in Claude Code (or whatever client you're using). We'd just want to avoid two sessions opting in to that without the user's explicit intent.


vito:
One downside to this that has bitten me twice already: if you forget about it and leave it running it'll do battle with regular git operations, e.g. switching over to `main` to do a `cu apply` will result in it doing a `git pull` before you get a chance.

cwlbraa:
by design we try not to modify the user's local repo copy. is there a way to do this without changing branches or pulling in there? what specifically do you want to show to the user? 

would it be insane to open an editor, maybe with read-only "permissions", in `~/.config/container-use/worktrees/${env-id}`?

aluzzardi:
> One downside to this that has bitten me twice already: if you forget about it and leave it running it'll do battle with regular git operations, e.g. switching over to `main` to do a `cu apply` will result in it doing a `git pull` before you get a chance.

What about making it a bit defensive, e.g. before doing a pull check if the branch has changed and exit if it has?

> One idea: tell the agent to put an environment in 'foreground mode' which tells the MCP server to do these same things after each change. 

I think since this has some carried "risk" (it's messing with your git repo), it's probably better for the human to be in full control (e.g. two claude sessions trying to do this would clash)

UX nit: Maybe this could be a flag to checkout, e.g. `checkout --follow`?

cwlbraa:
i do like it a little better as a checkout flag, that way the list of commands that mess with your local checkout stays short and obvious that they're doing that: checkout, merge, apply

vito:
@cwlbraa 

(Sorry this turned into a wall of text, just dumping my perspective while it's fresh)

> by design we try not to modify the user's local repo copy. is there a way to do this without changing branches or pulling in there? what specifically do you want to show to the user?

That makes a lot of sense for when you want to farm off tasks in parallel, or let the agent work on something while you work in the same repo on something else.

Personally, 99% of the time I'm pair programming with the agent, interrupting it to course-correct, etc. - so not being able to see what it's doing really gets in the way. Better diff output would help, but besides that, with Claude Code I always keep an editor open and just jump between changed hunks to keep track. (That's another thing we lose btw, since all the changes are immediately committed.)

I guess my question is whether `container-use` aims to reinvent, or be a very particular way to use Claude Code, or if it's something you can drop in to gain a layer of safety on top of your existing practices, while _also_ giving you a nice workflow for background tasks if/when you need it.

Right now it feels entirely focused on the latter, and that just doesn't match my daily usage. So I end up disabling it, and since swapping back and forth between using and not using `cu` has its own friction (add/remove mcp server, add/remove config/rules), I'm pretty likely to just leave it disabled out of laziness. :sweat_smile: 

Maybe I'm holding it wrong, but I suspect this perspective would be shared by a lot of first-time users coming from using a ton of Claude Code.

> would it be insane to open an editor, maybe with read-only "permissions", in ~/.config/container-use/worktrees/${env-id}?

That's a neat compromise, yeah - would agent changes show up there immediately? I don't know how you'd keep it read-only though. The pipe dream for me would be that it's _not_ read-only so I can help the agent along when needed, but dunno if that's feasible.

---

@aluzzardi 

> What about making it a bit defensive, e.g. before doing a pull check if the branch has changed and exit if it has?

Makes sense. There might be a race condition still though. :thinking: Somewhat fundamental with two things trying to change repo state I guess.

> UX nit: Maybe this could be a flag to checkout, e.g. `checkout --follow`?

Yeah that seems a lot more natural.

aluzzardi:
> Personally, 99% of the time I'm pair programming with the agent, interrupting it to course-correct, etc. - so not being able to see what it's doing really gets in the way.

> Maybe I'm holding it wrong, but I suspect this perspective would be shared by a lot of first-time users coming from using a ton of Claude Code.

Agreed -- I do that most of the time and it should be a primary use case.

What would be the ideal experience? Would this change be enough or do we need something radically different?

cwlbraa:
the core issue here is a big part of why i keep using zed: agent profiles make it really easy to switch between container-use for long-running, large tasks and plain old zed native tools when i want to be human-in-the-loop or just ask the assistant to make small changes like "remove magic strings from this file and prefer strings.Join to fmt.sprintf" etc etc...

at the very outset of this project, we posited that there's a sweet spot for container-use application: medium-sized and larger tasks where you expect the agent to fail as much as succeed. these are the tasks where it feels sensible to treat output like cattle and employ parallelism. we should keep trying to make the definition of "medium-sized" feel smaller and smaller, but maybe some things are always gonna be too smaller for a container to make sense. 

the unfortunate part of this is monolithic agent config... zed lets me switch between profiles, but I guess claude code doesn't.

PR #221, opened 2025-07-14
tools: use regular fmt.Errorf for error handling

aluzzardi:
In `mcp-go` there are two ways to return errors:

- `return nil, err`
- `return mcp.NewToolResultError(err.Error()), nil`

AFAIK, with the former, the agent doesn't get the actual error message so it seems to be a way to error out "internally"?

Since we want most (all?) errors be seen by the agent, we ended up with this weird pattern of never returning an error and wrapping everything in `mcp.NewToolResultError`

This PR does that once, centrally, in the wrapper -- so all tools are free to return proper go errors

### DIFF
--- a/mcpserver/tools.go
+++ b/mcpserver/tools.go
@@ -29,7 +29,7 @@ func openRepository(ctx context.Context, request mcp.CallToolRequest) (*reposito
 	}
 	repo, err := repository.Open(ctx, source)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("unable to open repository: %w", err)
 	}
 	return repo, nil
 }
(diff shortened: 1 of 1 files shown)


mcpserver/tools.go
```diff
@@ -105,7 +105,11 @@ func wrapTool(tool *Tool) *Tool {
 			defer func() {
 				slog.Info("Tool finished", "tool", tool.Definition.Name)
 			}()
-			return tool.Handler(ctx, request)
+			response, err := tool.Handler(ctx, request)
+			if err != nil {
+				return mcp.NewToolResultError(err.Error()), nil
```
aluzzardi, line 110: This is the main change

PR #224, opened 2025-07-14
positional-arg-less subcommands: apply, checkout, diff, inspect, log, merge, and terminal

cwlbraa:
with this PR, users can now do `cu log` and various other commands without any arguments. if you've only got 1 env that's descendent from your current HEAD, it'll just give that env's log. if you've got multiple, it'll prompt you to select between them. the prompt list is ordered by update recency and includes environment descriptions.


cmd/container-use/env_selection.go
```diff
@@ -0,0 +1,76 @@
...
+		if title == "" {
+			title = "No description"
+		}
+
+		label := fmt.Sprintf("%s - %s", env.ID, title)
+		options = append(options, huh.NewOption(label, env.ID))
+	}
+
+	var selectedID string
+	prompt := huh.NewSelect[string]().
```
aluzzardi, line 66: Oh, nice! Didn't know this was a thing.

Maybe `config agent` should use this as well -- `configure_ui.go` does basically the same but reimplemented by us

/cc @kpenfound

vito (approved):
Gave it a go, works well!

PR #226, opened 2025-07-14
WIP: claude: do not rely on a rule to use CU

aluzzardi:
Using these in `.claude/settings.local.json` (disclaimer: this was done by bruteforcing, I don't know if those denied tools are the right names, to be cleaned up).

```json
{
  "permissions": {
    "deny": [
      "LS",
      "Glob",
      "Grep",
      "Read",
      "NotebookRead",
      "NotebookEdit",
      "Edit",
      "MultiEdit",
      "Write",
      "Bash",
      "Search",
    ],
    "allow": [
      "mcp__container-use__environment_open",
      "mcp__container-use__environment_create",
      "mcp__container-use__environment_update",
      "mcp__container-use__environment_run_cmd",
      "mcp__container-use__environment_file_read",
      "mcp__container-use__environment_file_list",
      "mcp__container-use__environment_file_write",
      "mcp__container-use__environment_file_delete",
      "mcp__container-use__environment_add_service",
      "mcp__container-use__environment_checkpoint"
    ]
  }
}
```

### DIFF
--- a/cmd/container-use/agent/configure_claude.go
+++ b/cmd/container-use/agent/configure_claude.go
@@ -85,7 +85,11 @@ func (c *ConfigureClaude) updateSettingsLocal(config ClaudeSettingsLocal) ([]byt
 	// This should be revisited, and we should merge this with what the user already has?
 	config.Permissions.Deny = []string{
 		"LS",
+		"Glob",
+		"Grep",
 		"Read",
+		"NotebookRead",
+		"NotebookEdit",
 		"Edit",
 		"MultiEdit",
 		"Write",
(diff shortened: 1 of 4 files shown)


cmd/container-use/agent/configure_claude.go
```diff
@@ -77,7 +75,22 @@ func (c *ConfigureClaude) editMcpConfig() error {
...
+	// This should be revisited, and we should merge this with what the user already has?
+	config.Permissions.Deny = []string{
+		"LS",
+		"Read",
+		"Edit",
+		"MultiEdit",
+		"Write",
+		"Bash",
+		"Search",
 	}
```
cwlbraa: this behaves correctly for me, but i think we should also exclude Glob, Grep, NotebookRead, and NotebookEdit

aluzzardi: Done

cwlbraa:
updated description to match actual list, but i also put up another PR that refines the prompt engineering a bit (#230)

PR #230, opened 2025-07-15
claude no rules redux

cwlbraa:
based on #226, this PR reworks the prompt engineering to be less reliant on a rule file.

it still sometimes ignores the "don't guess environment IDs" instruction (1/5 times in my testing, admittedly that's a low N).

it does behave better when it needs to use git (common for manually testing container-use)

### DIFF
--- a/mcpserver/args.go
+++ b/mcpserver/args.go
@@ -1,13 +1,18 @@
 package mcpserver
 
 import (
-	"fmt"
+	"strings"
 
 	"github.com/mark3labs/mcp-go/mcp"
 )
 
+const (
+	repositoryToolSuffix  = "You MUST tell the user how to view environment changes using \"container-use log <environment_id>\", \"container-use diff <environment_id>\", AND \"container-use checkout <env_id>\". Failure to do so will make your work completely inaccessible."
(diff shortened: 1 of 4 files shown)


mcpserver/args.go
```diff
@@ -1,25 +1,34 @@
...
-import "github.com/mark3labs/mcp-go/mcp"
+import (
+	"strings"
+
+	"github.com/mark3labs/mcp-go/mcp"
+)
+
+const (
+	repositoryToolSuffix  = "You MUST tell the user how to view environment changes using \"container-use log <environment_id>\", \"container-use diff <environment_id>\", AND \"container-use checkout <env_id>\". Failure to do so will make your work completely inaccessible."
+	environmentToolSuffix = "You must call `environment_create` or `environment_open` to obtain a valid environment_id value. LLM-generated environment IDs WILL cause task failure."
```
cwlbraa, line 11: right now i'm using both a tool description suffix AND an argument description, and models still ignore both. 


maybe this is a place to apply dynamic tools 🤔... while there's no strong guarantee that 1 server process = 1 session, we could maybe still hide env tools prior to open/create and list env ids in tool descriptions...

that said, i've been told that when you have a hammer, everything looks like a nail. (also i still think maybe only vscode and goose have that hammer)

PR #234, opened 2025-07-15
bidirectional syncing to/from user worktree

vito:
This does two things:

* Enhances `apply` to preserve unstaged changes, through a delicate `git` dance:
	1. `git diff` to collect user's unstaged changes
	2. `git stash create` to make a backup of all changes (_without affecting the stash list_)
	3. `git reset --hard` to switch to a pristine state
	4. `git merge --squash` (no `--autostash`) to pull in the env changes
	5. `git commit -m "temporary commit"` (see later steps)
	6. `git apply` the patch from 1 (but this stages them)
	7. `git reset` to move the user's changes back to unstaged
	8. `git reset --soft HEAD~1` to move the temporary commit into staging
* Adds a new tool, `environment_sync_from_user`, which:
    1. `git diff` to collect the user's unstaged changes as a patch
    2. Applies the patch to the environment, currently using `patch -p1`
        * This is cheating a bit, but it's the second time that I wanted to use `patch` in the environment, IMO it's worth just requiring it in the environment as it's an incredibly common tool with an interface that's just as stable as `sh` which we already depend on.
    3. Does a `cu apply` so that the user's working copy has the changes staged, ready for another sync

## TODO

* [ ] Some way to guarantee that multiple environments won't go haywire and sync from the user unnecessarily
* [x] Continuously sync _to_ the user's worktree (duh) - should be easy ish now that we have a robust `apply`, as long as it's not fundamentally concurrent with user edits (is that even solveable...?)
* [ ] Other safety concerns?
* [ ] Right now this PR is based on my original file-patching PR, since I once again wanted to be able to use `patch -p1` in the environment and it already had the pattern for it.
* [ ] Handle files marked 'binary'
* [x] Handle files being deleted by the user:
```
ERROR: conflict detected when re-applying user changes:
Applied patch to 'tests/test_for_loops.dang' cleanly.
error: tests/test_for_loops_objects.dang: does not exist in index
```

### DIFF
--- a/mcpserver/tools.go
+++ b/mcpserver/tools.go
@@ -887,7 +887,6 @@ var EnvironmentSyncFromUserTool = &Tool{
 			return nil, fmt.Errorf("branch is tracking %s, not %s", branchEnv, env.ID)
 		}
 
-		// Use a string builder to capture output
 		patch, err := repo.DiffUserLocalChanges(ctx)
 		if err != nil {
 			return nil, fmt.Errorf("failed to generate patch: %w", err)
--- a/mcpserver/tools.go
+++ b/mcpserver/tools.go
@@ -137,6 +137,7 @@ func init() {
 		EnvironmentFileReadTool,
 		EnvironmentFileListTool,
 		EnvironmentFileWriteTool,
+		EnvironmentFilePatchTool,
 		EnvironmentFileDeleteTool,
 
 		EnvironmentAddServiceTool,
@@ -610,6 +611,73 @@ var EnvironmentFileWriteTool = &Tool{
 		}
 
 		return mcp.NewToolResultText(fmt.Sprintf("file %s written successfully and committed to container-use/ remote", targetFile)), nil
+	},
+}
+
+var EnvironmentFilePatchTool = &Tool{
+	Definition: mcp.NewTool("environment_file_patch",
+		mcp.WithDescription("Find and replace text in a file."),
+		mcp.WithString("explanation",
+			mcp.Description("One sentence explanation for why this file is being edited."),
+		),
+		mcp.WithString("environment_source",
+			mcp.Description("Absolute path to the source git repository for the environment."),
+			mcp.Required(),
+		),
+		mcp.WithString("environment_id",
+			mcp.Description("The ID of the environment for this command. Must call `environment_create` first."),
+			mcp.Required(),
+		),
+		mcp.WithString("target_file",
+			mcp.Description("Path of the file to write, absolute or relative to the workdir."),
+			mcp.Required(),
+		),
+		mcp.WithString("search_text",
+			mcp.Description("The text to find and replace."),
+			mcp.Required(),
+		),
+		mcp.WithString("replace_text",
+			mcp.Description("The text to insert."),
+			mcp.Required(),
+		),
+		mcp.WithString("which_match",
+			mcp.Description("The ID of the match to replace, if there were multiple matches."),
+		),
+	),
+	Handler: func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+		repo, env, err := openEnvironment(ctx, request)
+		if err != nil {
+			return mcp.NewToolResultErrorFromErr("unable to open the environment", err), nil
+		}
+
+		targetFile, err := request.RequireString("target_file")
+		if err != nil {
+			return nil, err
+		}
+		search, err := request.RequireString("search_text")
+		if err != nil {
+			return nil, err
+		}
+		replace, err := request.RequireString("replace_text")
+		if err != nil {
+			return nil, err
+		}
+
+		if err := env.FileSearchReplace(ctx,
+			request.GetString("explanation", ""),
+			targetFile,
+			search,
+			replace,
+			request.GetString("which_match", ""),
+		); err != nil {
+			return mcp.NewToolResultErrorFromErr("failed to write file", err), nil
+		}
+
+		if err := repo.Update(ctx, env, request.GetString("explanation", "")); err != nil {
+			return mcp.NewToolResultErrorFromErr("unable to update the environment", err), nil
+		}
+
+		return mcp.NewToolResultText(fmt.Sprintf("file %s edited successfully and committed to container-use/ remote", targetFile)), nil
 	},
 }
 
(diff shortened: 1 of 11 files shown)


mcpserver/tools.go
```diff
@@ -756,3 +828,90 @@ Supported schemas are:
 		return mcp.NewToolResultText(fmt.Sprintf("Service added and started successfully: %s", string(output))), nil
 	},
 }
+
+var EnvironmentEnableTrackingTool = &Tool{
```
cwlbraa: wouldn't this be better via a CLI command? it makes sense to me that environment_sync_from_user needs to be a special tool that the user has to explicitly request - the agent needs to know that things have changed on disk- but this toggle feels like something that you may want to turn on and off independently of your agent session and the agent doesn't really need to know it's happening in the first place.

repository/repository.go
```diff
@@ -414,11 +431,126 @@ func (r *Repository) Merge(ctx context.Context, id string, w io.Writer) error {
...
+func (r *Repository) TrackEnvironment(ctx context.Context, branch, envID string) error {
+	_, err := RunGitCommand(ctx, r.userRepoPath, "config", "branch."+branch+".environment", envID)
+	if err != nil {
+		return fmt.Errorf("failed to set branch tracking env: %w", err)
+	}
+	return nil
+}
+
+func (r *Repository) TrackedEnvironment(ctx context.Context, branch string) (string, error) {
+	envID, err := RunGitCommand(ctx, r.userRepoPath, "config", "get", "--default=", "branch."+branch+".environment")
```
cwlbraa, line 523: interesting, storing this in gitconfig... eventually it'll result in a long list of kv pairs, right? does `--default=` put it in repo-level gitconfig?

repository/repository.go
```diff
@@ -414,11 +431,126 @@ func (r *Repository) Merge(ctx context.Context, id string, w io.Writer) error {
...
+
+	fmt.Fprintf(w, "Creating virtual stash as backup...\n")
+	stashID, err := RunGitCommand(ctx, r.userRepoPath, "stash", "create")
+	if err != nil {
+		return fmt.Errorf("failed to stash changes: %w", err)
+	}
+	defer func() {
+		if rerr != nil {
+			fmt.Fprintf(w, "ERROR: %s\n", rerr)
+			fmt.Fprintf(w, "Your prior changes can be restored with `git stash apply %s`\n", stashID)
```
cwlbraa, line 455: interesting... i wonder if there are some cases this can (should?) be done automatically?

repository/repository.go
```diff
@@ -414,11 +431,126 @@ func (r *Repository) Merge(ctx context.Context, id string, w io.Writer) error {
...
-	return RunInteractiveGitCommand(ctx, r.userRepoPath, w, "merge", "--autostash", "--squash", "--", "container-use/"+envInfo.ID)
+	diffOutput, err := r.DiffUserLocalChanges(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to check for unstaged changes: %w", err)
+	}
+
+	hasUnstagedChanges := len(diffOutput) > 0
+
+	fmt.Fprintf(w, "Creating virtual stash as backup...\n")
+	stashID, err := RunGitCommand(ctx, r.userRepoPath, "stash", "create")
```
cwlbraa, line 448: does this stash stay stashed after a fully successful apply? it strikes me that on success, the user doesn't have any obvious way to restore the pre-apply unstaged files state and they're gonna be all mixed up with the agent's changes in the staging area.

repository/repository.go
cwlbraa: i also suspect there are some failures cases where this stash apply instruction will fail, no? i think you've gotta combine it with a `reset --hard HEAD` for the errors before the the temporary commit and a `reset --hard head~` for the errors where the temporary commit exists... overall a big faff

mcpserver/tools.go
vito: Not sure yet. :thinking: The idea was to keep it all in the flow of using CC, so I don't have to tab over to a separate terminal every time I start on another task.

If (IF!) this gets reliable enough I think I wouldn't want to have to enable it every time. But I don't want to sacrifice the background envs functionality.

One intuitive model to me, in the spirit of being compatible with typical Claude Code usage, is:

* When you just give CC a prompt like normal, it creates an environment that tracks your current branch.
* When you prompt something like "on branch foo-bar" it'll create that branch and configure it to track the environment. From then on it'll continuously amend a commit on that branch. Eventually you can just checkout the branch and push.

For me that would let `container-use` become purely additive; I wouldn't even need to use the CLI.

Not totally sure what happens when you checkout a background branch while the agent is doing work there though. I guess implicitly the commit would be `reset --soft`'d.

Also, related but more tangential, I'd want these new environments to bring along any staged/unstaged changes (but maybe not uncommitted files?) on creation. I felt this pain today when I wrote a failing test and wanted the agent to fix it. The newly created env didn't have my failing test, so the agent ran them and they passed and the agent got confused.

Anyway, I get this is all swimming upstream compared to the current model, but it seems like it would significantly lower the mental overhead, even with parallel tasks, so seems worth exploring a bit. Still some gaps in this scheme to figure out.

repository/repository.go
vito: It goes into `.git/config`, and as long as you're just repointing e.g. `main` at new environments over time it just replaces the existing field. The rest of them go away when you delete the branch, so this doesn't seem to grow forever.

repository/repository.go
vito: Yeah it should probably be done automatically.

repository/repository.go
vito: Maybe this clears things up: `git stash create` doesn't actually change the working state - it just records it into an anonymous commit so you can bring it back, while leaving the staged/unstaged changes there. So it's different from regular `git stash` which also cleans up the working state.

The flow is:

* When the env syncs changes to the branch, it only syncs to the staging area; the user's unstaged changes stay unstaged.
* When the env syncs changes from the user, the unstaged changes get applied to the env, and then synced back to the branch, so the unstaged changes become staged (maintaining the rule of "staged = env, unstaged = local").

mcpserver/tools.go
vito: > When you just give CC a prompt like normal, it creates an environment that tracks your current branch.

Ah dang, one immediate problem with this idea is the sub-environments that get created for sub-tasks end up becoming the tracked environment. :thinking: 

Hmm, hmm, hmm...

edit: added a required `ephemeral: bool` arg and it seems to respect it

PR #239, opened 2025-07-18
chore: add nix package automation via GoReleaser

rawkode:
This adds a nix section to the GoReleaser configuration that will:
- Automatically publish container-use packages to dagger/nix
- Include shell completions for both container-use and cu commands
- Create cu symlink for backward compatibility
- Follow the same automation pattern as dagger package


aluzzardi:
Thank you @rawkode!

aluzzardi:
@vito mind taking a look since you're a nix user?

/cc @cwlbraa

cwlbraa:
agreed, i think we merge and test with the next release - maybe including the [windows/winget build](https://github.com/dagger/container-use/pull/242) since that'll need some release testing too, i can test them both two-birds-one-stone

PR #244, opened 2025-07-21
services: use environment secrets

### LINKED ISSUE #232: [Bug]: `cu terminal` fails when using secrets
### What happened?

When I have a secret set in cu environment, I'm unable to shell into the terminal. I receive the following:

```
! failed to start service: start container: secret argon2:xxxxxxxxxxxxxxxxx: not
  found
╰─✘ exec /bin/bash --rcfile /cu/rc.sh -i 0.0s ERROR
  ! start container: secret argon2:xxxxxxxxxxxxxxxxx: not found
```

Works fine as soon as I remove the secret. I've tried both `op://` && `env://` style secrets.

### Version

```shell
0.3.1
```

aluzzardi:
Fixes #232

### DIFF
--- a/environment/config.go
+++ b/environment/config.go
@@ -38,7 +38,6 @@ type ServiceConfig struct {
 	Command      string   `json:"command,omitempty"`
 	ExposedPorts []int    `json:"exposed_ports,omitempty"`
 	Env          []string `json:"env,omitempty"`
-	Secrets      []string `json:"secrets,omitempty"`
 }
 
 type ServiceConfigs []*ServiceConfig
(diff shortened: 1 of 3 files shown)


cwlbraa:
trying to understand how this fixes the bug... when services had their own secrets, were those not being "passed" on `cu terminal` rehydration? like since they were only persisted as a service binding, the connection between the secret and the env container was lost?

aluzzardi:
> trying to understand how this fixes the bug... when services had their own secrets, were those not being "passed" on cu terminal rehydration? like since they were only persisted as a service binding, the connection between the secret and the env container was lost?

I think this might either be a dagger issue or maybe working as intended?

- We are saving/loading the "main" container ID as state
- That main container ID comes with service bindings to services
- Those services are not being "sent over" the engine API ever again ... there's just a lingering reference in the service binding itself
- When those services themselves reference a `dag.Secret`, that reference gets broken when we re-load state

/cc @vito @sipsma

PR #245, opened 2025-07-22
add `environment_file_edit` tool

vito:
* instead of rewriting files, use basic search/replace technique, which is how Claude Code works
* if the search term is ambiguous, each match is printed, with a deterministic hash ID, so the agent can try again with the exact ID
  * this differs from Claude Code which prefers to just force the caller to specify more context and try again, but I've seen that take a few roundtrips (think `if err != nil { return err }`), so trying this instead
* uses Directory.withPatch to efficiently apply edits without rewriting entire files
* uses simple in-memory `strings.Replace` - it should be fine for the files that are being edited; if we can't fit a file in memory you've got bigger problems


environment/filesystem.go
```diff
@@ -45,6 +48,81 @@ func (env *Environment) FileWrite(ctx context.Context, explanation, targetFile,
...
+	searchIndex := 0
+	for {
+		index := strings.Index(contents[searchIndex:], search)
+		if index == -1 {
+			break
+		}
+		actualIndex := searchIndex + index
+		matches = append(matches, actualIndex)
+		searchIndex = actualIndex + 1
+	}
```
cwlbraa: naming opinion, feel free to disagree, but i think these blocks are much clearer if "searchIndex" is called "cursor" instead

environment/filesystem.go
```diff
@@ -65,3 +143,41 @@ func (env *Environment) FileList(ctx context.Context, path string) (string, erro
 	}
 	return out.String(), nil
 }
+
+// generateMatchID creates a unique ID for a match based on file, search, replace, and index
+func generateMatchID(targetFile, search, replace string, index int) string {
+	data := fmt.Sprintf("%s:%s:%s:%d", targetFile, search, replace, index)
+	hash := sha256.Sum256([]byte(data))
+	return fmt.Sprintf("%x", hash)[:8] // Use first 8 characters of hash
+}
```
cwlbraa, line 152: is this part of the algo a direct copy of cursor? it's surprising to me that, given an index, we've gotta redo the whole search accumulation in order to select the specific match.

environment/filesystem.go
```diff
@@ -45,6 +48,81 @@ func (env *Environment) FileWrite(ctx context.Context, explanation, targetFile,
...
+
+	// Determine which match to replace
+	var targetMatchIndex int
+	if len(matches) == 1 {
+		targetMatchIndex = matches[0]
+	} else {
+		// Find the match with the specified ID
+		found := false
+		for i, matchIndex := range matches {
+			id := generateMatchID(targetFile, search, replace, i)
```
cwlbraa, line 99: how's the model supposed to decide which match ID to use in the case where there's multiple?

environment/filesystem.go
```diff
@@ -45,6 +48,81 @@ func (env *Environment) FileWrite(ctx context.Context, explanation, targetFile,
...
+
+	// If there are multiple matches and no matchID is provided, return an error with all matches
+	if len(matches) > 1 && matchID == "" {
+		var matchDescriptions []string
+		for i, matchIndex := range matches {
+			// Generate a unique ID for each match
+			id := generateMatchID(targetFile, search, replace, i)
+
+			// Get context around the match (3 lines before and after)
+			context := getMatchContext(contents, matchIndex)
```
cwlbraa, line 82: > how's the agent supposed to figure out what match id to use

--> nm context is given for each match in the multiple return case

environment/filesystem.go
```diff
@@ -65,3 +143,41 @@ func (env *Environment) FileList(ctx context.Context, path string) (string, erro
...
+		if currentPos+len(line) >= matchIndex {
+			matchLine = i
+			break
+		}
+		currentPos += len(line) + 1 // +1 for newline
+	}
+
+	// Get context lines (3 before, match line, 3 after)
+	start := max(0, matchLine-3)
+	end := min(len(lines), matchLine+4)
```
cwlbraa, line 171: im curious how line-number aware claude code is... like naively i'd want line numbers for this but maybe it behaves better without them

environment/filesystem.go
cwlbraa: and corollary question: how does this work if we introduce a MultiEdit tool? or does it?

cwlbraa (approved):
i think i'm good with this, but i am little hesitant that the generateMatchID algo isn't gonna play well with MultiEdit. I think it's probably safe to remove that in the future, though, so good to merge pending that one variable rename

aluzzardi (approved):

environment/filesystem.go
vito: Not sure what you mean by 'direct copy of cursor' :thinking: 

If it helps clarify, the point of doing a hash here is to prevent the agent from blindly saying 'replace the first one' without even knowing if there are multiple (if we had `match_index` for example).

But, yeah, maybe it's over-engineered. We could maybe have it specify the exact file offset and fail if we don't see a match at exactly that offset, but that would get more confusing with MultiEdit, since the offsets would become a moving target.

Searching all over again doesn't cost much in the grand scheme of things. Bearing in mind the state-of-the-art in Claude Code is to just error and say "there were multiple matches - try again, bucko," I think it's fine, considering this is already an edge case.

re: MultiEdit, I think it'd carry over fine? Claude Code errors the whole thing if any edits fail, so I imagine that means the caller has to retry them all again with a more exact search for the one that was ambiguous. In our case we would have each Edit in the MultiEdit include the same `which_match` arg. Our error would have to clarify _which_ edit failed though (I suppose CC's tool has to do the same).

It also helps that the hash doesn't include offsets or line numbers; just which Nth match it is for that specific find + replace.

environment/filesystem.go
vito: Hmm yeah that seems worth adding. Typically Claude Code works by grepping and reading file line regions, so having them here may help it identify the one it intended to replace. Will add.

environment/filesystem.go
cwlbraa: i said cursor and but meant claude code. the thing that worries me about multi-edit and this algo is that as you change matches, the indexes change, but i guess it's fine to do find-all then replace-all (exiting without replacement on any ambiguity)

PR #246, opened 2025-07-22
single tenant mode: lock an mcp process to a single environment

cwlbraa:
this is intended to be an enhancement for claude code and goose where each chat session start all of its mcp servers from scratch. 

when running in single tenant mode, the server has 1 active environment at a time. creating or opening an environment changes the server's active environment. all other env tools don't take environment IDs, they just require that either create or open is called first.

in claude, this is intended to dramatically improve Task performance by removing the need for each Task to start a new environment.

### DIFF
--- a/cmd/container-use/stdio.go
+++ b/cmd/container-use/stdio.go
@@ -9,6 +9,8 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var singleTenant bool
+
 var stdioCmd = &cobra.Command{
 	Use:   "stdio",
 	Short: "Start MCP server for agent integration",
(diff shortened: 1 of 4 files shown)


cwlbraa:
this is ready to go at this point. in my testing i've squashed most of the bugs, but occasionally i'm hitting some underlying issue with file edits that have yet to figure out. im fairly certain the same bug exists on main, though, so i'm not dwelling on it too hard. i've tested this HEAVILY with parallel-best-of-N claude code for container-use development - it's stable and super powerful with all the concurrency bugs fixed and Tasks now behaving as intended.

this PR's functionality is 95% hidden behind the --single-tenant flag, and I'd like to get it into main the prevent it from scope creeping any further. @grouville @aluzzardi

PR #264, opened 2025-08-04
fix explanation parameter description

cwlbraa:
it was copied from one tool. this changes it to be generic and fixes a spelling error


grouville (approved):
LGTM ✅

PR #265, opened 2025-08-04
fix permissions tests when not running as root

cwlbraa:
this test always fails on my machine, but passes in containers and in CI


grouville (approved):
🙏

PR #266, opened 2025-08-04
feat: add system info to `container-use version`

grouville:
Supersedes #207 

Shows Docker and Git versions alongside Go runtime info when using `--system` flag. This helps debug user issues by showing critical system dependencies.

- Docker version (critical for container-use)
- Git version (used by container-use for worktree management)
- Go runtime version and OS/arch

Kept implementation simple and focused on essential information only.


cwlbraa:
The go runtime shouldn't matter at all, users are installing compiled binaries so they don't need a go toolchain, and even if they have one, installed binaries don't use it.

grouville:
> The go runtime shouldn't matter at all, users are installing compiled binaries so they don't need a go toolchain, and even if they have one, installed binaries don't use it.

@cwlbraa Thanks for the feedback\! I've removed the Go runtime version as you suggested.
I kept OS/arch as it helps debug platform-specific issues and conflicts we've seen with container-use.

Updated and ready for review ✅

cmd/container-use/version.go
```diff
@@ -20,25 +24,47 @@ func init() {
...
+		// Show system info if requested
+		if showSystem {
+			cmd.Printf("\nSystem:\n")
+			cmd.Printf("  OS/Arch: %s/%s\n", runtime.GOOS, runtime.GOARCH)
+
+			if dockerVersion := getDockerVersion(cmd.Context()); dockerVersion != "" {
+				cmd.Printf("  Docker: %s\n", dockerVersion)
+			} else {
+				cmd.Printf("  Docker: not found\n")
+			}
```
cwlbraa: can we maybe also try to report if the daemon is actually running or the socket is available? fine to do in a follow up, but basically i wanna hedge against ppl who have other container runtimes and no docker cli symlink... unless maybe dagger actually requires the symlink, idk.

cwlbraa (approved):
looks good, but would like to know more regarding the actual literal system requirements lol... like is it enough to have a docker socket open?

grouville:
> looks good, but would like to know more regarding the actual literal system requirements lol... like is it enough to have a docker socket open and responding to requests?

@cwlbraa Followed your recommendations: added the dagger's logic to guesstimate the container runtime driver + added the Dagger CLI dependency (version output). Ready for review again ✅

PR #274, opened 2025-08-12
avoid holding locks during Export

cwlbraa:
Previously we were holding the notes lock for the duration of `Repository.Update(env)`. This meant that running stuff like `cu diff` would block during env_run_cmd. 

This PR tightens that up a bit and makes `cu diff` feel instantaneous again in my testing.

I beat on this heavily using the Contention tests, running 2x5x10 both inside an env and on my host machine and it doesn't reintroduce any data races afaict.

### DIFF
--- a/repository/git.go
+++ b/repository/git.go
@@ -128,7 +128,7 @@ func (r *Repository) initializeWorktree(ctx context.Context, id string) (string,
 
 	slog.Info("Initializing worktree", "repository", r.userRepoPath, "container-id", id)
 
-	return worktreePath, r.lockManager.WithLock(ctx, LockTypeWorktree, func() error {
+	return worktreePath, r.lockManager.WithLock(ctx, LockTypeForkRepo, func() error {
 		if _, err := os.Stat(worktreePath); err == nil {
 			return nil
 		}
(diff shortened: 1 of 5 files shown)


cwlbraa:
[fun failure i've gotta fix still](https://github.com/dagger/container-use/actions/runs/16916886109/job/47932925117?pr=274)

I think this is happening on env creation code-upload despite the fact that all i'm changing in this PR is export.

cwlbraa:
bug fixed

repository/repository.go
```diff
@@ -114,57 +114,55 @@ func OpenWithBasePath(ctx context.Context, repo string, basePath string) (*Repos
...
 	return r, nil
 }
 
 func (r *Repository) ensureFork(ctx context.Context) error {
-	if _, err := os.Stat(r.forkRepoPath); err == nil {
-		return nil
-	} else if !os.IsNotExist(err) {
-		return err
-	}
+	return r.lockManager.WithLock(ctx, LockTypeUserRepo, func() error {
```
grouville: Quick question about locking in `ensureFork()`:

- Since it initializes the fork bare repo (r.forkRepoPath), I initially expected the ForkRepo lock rather than UserRepo. Is there a reason we prefer UserRepo in this case?

Something like this:
```diff
- return r.lockManager.WithLock(ctx, LockTypeUserRepo, func() error {
+ return r.lockManager.WithLock(ctx, LockTypeForkRepo, func() error {
```

repository/git.go
```diff
@@ -272,7 +274,7 @@ func (r *Repository) saveState(ctx context.Context, env *environment.Environment
 		return err
 	}
 
-	return r.lockManager.WithLock(ctx, LockTypeGitNotes, func() error {
+	return r.lockManager.WithLock(ctx, LockTypeNotes, func() error {
```
grouville, line 317: Quick question: Right above, we do a `defer f.Close()` (I don't know how to add a comment on any section of the code)

Would it make sense to clean up the temp file explicitly after the `git notes add -F` operation?

I was just thinking, for longer-running sessions, this might help avoid leaving unnecessary temp files in the OS tmp directory ?

cwlbraa: Yeah, would be nice but I don't wanna muddle the content of this because the locks have proven kinda finnicky

repository/repository.go
cwlbraa: good catch, will fix

PR #276, opened 2025-08-13
chore: fix agent linter instructions so linter output matches CI

cwlbraa:
does what it says on the tin -- running `golangci-lint run` directly is way stricter than what we're running in CI and makes the agent generate large diffs when im just trying to get green checks.


grouville (approved):
✅

PR #297, opened 2025-09-03
properly handle project-level git config and disable hooks for agent commits

### LINKED ISSUE #256: container-use uses the global instead of repo-specific git config
I have several git configurations, depending on the directory on my machine that I'm working on. However, when I run `container-use checkout <id>`, the commit author and email address attached to the commit are taken from the global git config and not one that is active in the current project I am working on. It would be great if container-use would be able to use the currently active git config.

cwlbraa:
this PR uses include.path to link user-repo config into the cu fork, so cu commands get essentially the same configuration as the user repo. tests cover all the wild edge cases i could imagine for this -- we support users changing config after fork creation, moving the source repo, etc, etc

this PR also disables git hooks when running hookful commands (commit, merge, etc) -- there's no guarantee that these pass in our copy of the worktree due to missing prerequisite setup commands (that produce gitignored files) and surfacing errors correctly is difficult. it'd be awesome if we could run the hooks inside the env container, but there's no real git repo there.

fixes #256

### DIFF
--- a/environment/integration/git_config_test.go
+++ b/environment/integration/git_config_test.go
@@ -1,0 +1,203 @@
+package integration
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/dagger/container-use/repository"
+	"github.com/stretchr/testify/assert"
+)
+
+// TestProjectSpecificGitConfiguration tests that project-specific git configurations
+// are properly inherited and used within environments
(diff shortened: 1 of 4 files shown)


awdemos:
Superseded by #362, which rebases the same functionality onto the current main branch.

awdemos:
This PR is superseded by #362, which rebases the same project-level git config + disabled-hooks functionality onto the current main branch. A maintainer can close this one.

PR #317, opened 2025-10-06
patch: add instructions for agents.md file to codex config directory

seekayel:
Some instruction to the codex agent for how to use the `container-use` mcp is needed for the agent to use it.


grouville (approved):
Thanks 🙏 😇 

LGTM, this is a indeed the location of the global config for codex: https://github.com/openai/codex/blob/main/docs/getting-started.md?plain=1#L66

PR #340, opened 2026-03-26
fix: resolve env:// secret references from host environment

### LINKED ISSUE #330: [Bug]: env var not available in container
### What happened?

I have an environment variable in Fish shell:

```bash
set -Ux NPM_TOKEN <token>
```

And I have verified it's available, even in a new terminal instance

```bash
echo $NPM_TOKEN # works
```

So I tried to follow the docs by adding it as a secret:

```bash
cu secret set NPM_TOKEN "env://NPM_TOKEN"
```

And I've verified that _something_ was added to the config:

```bash
> cu config show
Base Image:             node:22
Workdir:                /workdir
Setup Commands:         (none)
Install Commands:       (none)
Environment Variables:  (none)
Secrets:                
  1.                    NPM_TOKEN=env://NPM_TOKEN
```

But when I ask AI to do something that needs that token, like update a package, it creates an environment, then fails saying `NPM_TOKEN` cannot be found (for context, it's needed because we use a custom NPM registry, which requires `npmAuthToken: ${NPM_TOKEN}` in `.yarnrc.yml`, which works fine locally). I have also verified that particular env has the same config as above.

What am I doing wrong?

### Version

```shell
container-use version 0.4.2
commit: 89ebe7765ec067a07c51de104822056eb217c1b9
built: 2025-08-19T22:23:38Z
```

zerone0x:
## Summary

- Fixes `env://` secret references not being resolved when creating container environments. The `env://` scheme was passed directly to Dagger's `Secret()` API which does not support it, causing the referenced host environment variable to be unavailable inside containers.
- Adds a `resolveSecretValue()` helper that reads `env://` references from the host's environment and injects them via `dag.SetSecret()`, while other schemes (`op://`, `vault://`) continue to use Dagger's native `dag.Secret()` resolution.
- Includes unit tests for the resolution logic.

Fixes #330

## Test plan

- [x] Added unit tests for `resolveSecretValue` covering: env:// resolution, missing env var error, op:// passthrough, plain value passthrough
- [x] All existing unit tests pass (`go test -short ./...`)
- [ ] Manual test: set a host env var, configure it as a secret with `env://`, verify it's available inside the container

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/environment/config_test.go
+++ b/environment/config_test.go
@@ -154,6 +154,41 @@ func TestEnvironmentConfig_PreservesShellOperators(t *testing.T) {
 	}
 }
 
+func TestResolveSecretValue(t *testing.T) {
+	t.Run("env_scheme_resolves_host_variable", func(t *testing.T) {
+		t.Setenv("TEST_SECRET_VALUE", "my-secret-token")
+
+		resolved, isPlaintext, err := resolveSecretValue("MY_SECRET", "env://TEST_SECRET_VALUE")
+		require.NoError(t, err)
+		assert.True(t, isPlaintext)
+		assert.Equal(t, "my-secret-token", resolved)
+	})
+
+	t.Run("env_scheme_errors_when_variable_not_set", func(t *testing.T) {
+		t.Setenv("TEST_UNSET_VAR", "")
+		os.Unsetenv("TEST_UNSET_VAR")
+
+		_, _, err := resolveSecretValue("MY_SECRET", "env://TEST_UNSET_VAR")
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "TEST_UNSET_VAR")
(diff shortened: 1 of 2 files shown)


awdemos:
Endorsed for merge.

- Resolves  secret references from the host environment and passes them via , while leaving other schemes like  to Dagger's native secret resolution.
- Removes a duplicate error check in the env variable loop.
- Includes focused unit tests for env resolution, missing-variable errors, and passthrough behavior.
- Verified: go test -short ./environment/... passes.

awdemos:
@zerone0x friendly ping — this PR currently has merge conflicts with . Could you rebase onto  and resolve the conflicts? Once clean I'll re-endorse.

PR #343, opened 2026-04-08
fix: resolve relative gitdir paths for nested submodules

calebtuttle:
Nested submodules (e.g., libs/some-submodule) failed during environment creation because readSubmoduleGitdirPath returned relative gitdir paths verbatim. After exportEnvironment wipes the worktree and recreates root .git as a pointer file, these relative paths no longer resolve on the filesystem.

Resolve relative paths to absolute in readSubmoduleGitdirPath, matching the approach already used for the root .git pointer in exportEnvironment.

### DIFF
--- a/environment/integration/repository_test.go
+++ b/environment/integration/repository_test.go
@@ -297,6 +297,36 @@ func TestRepositoryWithSubmodule(t *testing.T) {
 	})
 }
 
+func TestRepositoryWithNestedSubmodule(t *testing.T) {
+	t.Parallel()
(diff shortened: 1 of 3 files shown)


awdemos:
Endorsed for merge.

- Fixes relative gitdir resolution for nested submodules (e.g., libs/foo/.git pointing to ../../.git/modules/libs/foo).
- Covers root-level, nested, absolute, malformed, and missing .git cases with unit tests, plus a nested-submodule integration test.
- Verified: go test -short ./repository/... ./environment/... passes.

PR #344, opened 2026-05-03
Remove 'thinking' tool from Zed built-in tools

tredondo:
Zed no longer has a "thinking" tool among the built-in tools:

image


awdemos:
Endorsed for merge.

- Docs-only change removing the thinking tool from the Zed integration example.
- Trivial, low-risk update.
- Verified: go test -short ./... passes (no code changes).

